PR #16, opened 2021-11-06
Feature: Add more flexible models to case study

seabbs:
This PR adds more flexible models to the case study. It also adds more explanation to both the readme and the case study covering what a nowcast is and what form the input data needs to be in.


seabbs:
LGTM (self-review)

PR #18, opened 2021-11-08
Bug: drop na.rm requirement for posterior functions

seabbs:
This PR drops `na.rm` requirement from some posterior package


seabbs:
LGTM (self-review)

PR #19, opened 2021-11-16
Feature: @sbfnk comments

seabbs:
This PR adds changes based on recent comments from @sbfnk. These changes are:

- Update the hazard to probabiltiy and probability to hazard equations to use improved notation
- Adding text explaining the use of `enw_preprocess_data()`
- Adding text giving more context for the posterior plots
- Adding additional text highlighting that the approximate leave one out information criterion is not a principled measure for this use case and providing a link out to the approximate leave future out cross validation vignette. 
- Adding text to explain the logit hazard step in more detail and to clarify the statement that the final hazard is 1.


seabbs:
LGTM (self-review)

PR #20, opened 2021-11-17
Bug: Fix presence of duplicate summarise samples

seabbs:
This PR fixes a bug which led to duplicated summarised samples when samples were summarised manually vs using `cmdstanr`.


seabbs:
LGTM (self-review)

Issue #7, opened 2021-11-04
Add true formula interface


seabbs:
At the moment the formula interface is flexible but not nice to use. Adding a true formula interface would improve the workflow with the package. This needs: 

- [x] Support fixed effects by passing directly `model.matrix`
- [x] Support random effects (passing `(1 + x | y)` correctly)
- [x] Support random walks (likely need to add a `rw` helper and extract this from the formula
- [x] Continue to support the current interface as likely above changes will not support all possible models.

seabbs:
A draft version of this is now implemented in #27 though documentation, testing, and fixing edge cases is still needed.

seabbs:
Done in #27

PR #27, opened 2022-01-29
Feature: Add formula interface

seabbs:
This PR contains work towards a full formula interface (i.e #7). 

There is a working formula interface that supports fixed effects, random effects (in the style of `lme4`), and random walks. I have updated the README to reflect this new interface. 

TODOs

- [x] Update the case study vignette to support this new interface
     - [x]  Update formulas and check design matrices are as previously specified (https://github.com/epiforecasts/epinowcast/pull/27/commits/354edd06f56eca4f33fbe56bf27a07178c913ce4)
     - [x] Check simple models can be fit and return expected results
     - [x] Check complex models can be fit and return expected results
     - [x] Rerender case study vignette 
- [x] Fix edge case where no report date model is desired (https://github.com/epiforecasts/epinowcast/pull/27/commits/67df4473a44f2debfe7b3e6a7a4641be4b4c24dd).
    - This is cut set up such that a model with an intercept only is equivalent to no model (as the intercept is set to 0). This is perhaps unclear. 
- [x] Decide if the formula interface should automatically add an intercept (i.e `~ 1`) this is required for some functionality to work at the moment but may not always be desirable (though not yet clear in what scenarios this is true). 
    - Current plan is to handle this instead in the interface update vs in this PR. 
- [x] Document all newly added functions
- [x] Add examples for newly added functions
- [x] Add tests based on these examples
- [x] Identify critical edge cases, fix, and add tests
- [x] Update version and news
- [x] Add a develop branch to merge this feature into to avoid many iterations of breaking changes on main.
- [x] Explore the implications of the random walk implementation. Is it acceptable that it is only a random walk in the residuals and not across all potential effects (i.e it does not include linear effect terms in previous weeks when by week).


seabbs (commented):
LGTM (self-review)

PR #38, opened 2022-04-05
Vectorised truncation distributions

seabbs:
This PR changes all truncated distributions that are vectorised to use an explicit form of the truncation distribution based on this discussion: https://discourse.mc-stan.org/t/truncation-with-vectorized-sampling-statements/20084/4

Testing was performed on the README example. No apparent change in either posterior estimates or overall model performance. It is likely more testing is needed on larger models to see any difference in performance (as no actual advantage of vectorisation in this simple example).


seabbs:
Running the first two models in the vignette and exploring I again see no impact on posterior estimates. However, I do see a reversion in performance with the first model taking ~380 seconds and the second ~840 seconds. It seems difficult to understand how this PR could be causing this so checking the code on master in the same computational environment.

seabbs:
The same two models using the code in `main` take ~360 seconds and ~820 seconds. The performance regression is likely a factor of the compute environment given these results (my battery-powered laptop vs desktop). However, there is still little to no evidence of improved performance with vectorised truncation so this PR is largely optional depending on if the code implemented in this way is clearer and if larger models are more likely to show performance improvements.

inst/stan/epinowcast.stan
```diff
@@ -125,8 +125,9 @@ transformed parameters{
 model {
   // priors for unobserved expected reported cases
   leobs_init ~ normal(eobs_init, 1);
+  eobs_lsd ~ normal(eobs_lsd_p[1], eobs_lsd_p[2]);
+  target += -normal_lccdf(0 | eobs_lsd_p[1], eobs_lsd_p[2]);
```
adrian-lison, line 129: Did you maybe forget to multiply the `-normal_lccdf` part with the number of groups `g`? I think the complementary cdf must be subtracted once for each "observation".

inst/stan/epinowcast.stan
```diff
@@ -137,19 +138,18 @@ model {
...
     logsd_eff ~ std_normal();
     if (neff_sds) {
-      for (i in 1:neff_sds) {
-        logmean_sd[i] ~ normal(logmean_sd_p[1], logmean_sd_p[2]) T[0,];
-        logsd_sd[i] ~ normal(logsd_sd_p[1], logsd_sd_p[2]) T[0,];
-      }
+      logmean_sd ~ normal(logmean_sd_p[1], logmean_sd_p[2]);
+      target += -normal_lccdf(0 | logmean_sd_p[1], logmean_sd_p[2]);
+      logsd_sd ~ normal(logsd_sd_p[1], logsd_sd_p[2]);
+      target += -normal_lccdf(0 | logsd_sd_p[1], logsd_sd_p[2]);
```
adrian-lison, line 144: Same here, `normal_lccdf` should probably be multiplied by `neff_sds`.

inst/stan/epinowcast.stan
```diff
@@ -137,19 +138,18 @@ model {
...
   }
   // priors and scaling for date of report effects
   if (nrd_effs) {
     rd_eff ~ std_normal();
     if (nrd_eff_sds) {
-      for (i in 1:nrd_eff_sds) {
-        rd_eff_sd[i] ~ normal(rd_eff_sd_p[1], rd_eff_sd_p[2]) T[0,];
-      }
+      rd_eff_sd ~ normal(rd_eff_sd_p[1], rd_eff_sd_p[2]);
+      target += -normal_lccdf(0 | rd_eff_sd_p[1], rd_eff_sd_p[2]);
```
adrian-lison, line 152: Multiply by `nrd_eff_sds`

inst/stan/epinowcast.stan
adrian-lison, line 129: Generally, I like vectorization not only because of efficiency but also because it makes the code more concise. In the present case, the two-liner is a bit unpleasant. It could be reduced to a one-liner as proposed by bbbales2 in the stan forum discussion:

`target += normal_lpdf(eobs_lsd | eobs_lsd_p[1], eobs_lsd_p[2]) - normal_lccdf(0 | eobs_lsd_p[1], eobs_lsd_p[2]) * g;`

This is not great but at least indicates that this is one likelihood component that belongs together. If we want to take it even further, we could write a user-defined function for the vectorized truncated normal distribution (until its available in stan itself). This would be the most readable version I guess!

adrian-lison (changes requested):
Nice idea, I added some comment on a missing factor in the subtracted log-likelihood below.

Also, to be honest, I haven't fully wrapped my head around the whole truncation, because I once learned that a half-normal prior can for example just be encoded by adding bounds to the parameter in the parameters block (and then proposals which violate the bounds will get rejected). Is the truncation of the distributions then only an efficiency improvement (less proposals needed) or does that really make a difference in terms of the model?

inst/stan/epinowcast.stan
seabbs: Yes, I agree. I like the idea of defining a custom vectorised truncated normal. Will add that in. Would help with mistakes as well (like missing all the multiplying by group...)

seabbs:
> Also, to be honest, I haven't fully wrapped my head around the whole truncation, because I once learned that a half-normal prior can for example just be encoded by adding bounds to the parameter in the parameters block (and then proposals which violate the bounds will get rejected). Is the truncation of the distributions then only an efficiency improvement (less proposals needed) or does that really make a difference in terms of the model?

I think it is just making the prior be "correct" and hence improve efficiency though I haven't seen that much discussion on it so its a bit of an informed guess.

adrian-lison:
No, I like doing it the more "correct" way - especially if we do it via a user-defined function! Then it will be very clean.

seabbs:
I'm pretty happy with this so keen to merge. @adrian-lison no pressure to review as I realise we have speeding into a holder season but let me know if happy to let this one into `main` :)

adrian-lison (dismissed):
Now that's some beautiful code! :)

seabbs:
> I'm pretty happy with this so keen to merge. @adrian-lison no pressure to review as I realise we have speeding into a holder season but let me know if happy to let this one into main :)

Sometimes I wonder if English can actually be my first language. What even happened to this sentence.

seabbs:
Good idea to add the error handling and thanks for the typo clean up :)

inst/stan/functions/zero_truncated_normal.stan
```diff
@@ -0,0 +1,13 @@
+real zero_truncated_normal_lpdf(vector y, real mu, real sigma) {
```
seabbs, line 1: I've been reading more about this and I think the lower bound only version might actually be functionality equivalent. Still not 💯  but pretty sure.

seabbs: I don't think it hurts to have it explicitly implemented but could potentially be fractionally more performant to just use a lower bound.

adrian-lison: Let's stay with the explicit truncation for now, as discussed :)

adrian-lison (approved):
Ready 2 go

PR #40, opened 2022-04-06
Feature vectorise likelihood

seabbs:
This PR adds a crude approach for fully vectorising the likelihood. It does so by first flattening the reporting triangle into a vector and creating the required index look ups to track which snapshot is linked to which observations in this vector. 

In the stan code this is the used in the `obs_lmf` function to assign observations in that parallel chunk as well as assigning expected observations from each snapshot into a comparable vector. No other changes are made meaning that the generated quantities are still using the rowwise version of th code (though these are called far less often so in theory optimising them should matter far less),.

These created objects are then called when calculating the log-likelihood outside of the snapshot loop (rather than doing this additively for each iteration). 

Initial testing suggests this has no impact on the results. Sense checking the function return indicates everything is working as intended. Running the readme example indicates a slight performance reversion with fit times dropping to ~ 100 seconds vs ~90 seconds for the code on main. Testing the first two models in the vignette suggests approximately no change for either model (+/- 30 seconds).

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -266,13 +266,13 @@ enw_model <- function(model, include, compile = TRUE,
     include <- system.file("stan", package = "epinowcast")
   }
 
-  if (compile) {
-    if (!profile) {
-      code <- paste(readLines(model), collapse = "\n")
-      code_no_profile <- remove_profiling(code)
-      model <- cmdstanr::write_stan_file(code_no_profile)
-    }
+  if (!profile) {
+    code <- paste(readLines(model), collapse = "\n")
+    code_no_profile <- remove_profiling(code)
+    model <- cmdstanr::write_stan_file(code_no_profile)
+  }
 
+  if (compile) {
     if (verbose) {
       model <- cmdstanr::cmdstan_model(model,
         include_paths = include,
@@ -302,9 +302,9 @@ enw_model <- function(model, include, compile = TRUE,
 #' @param s Character vector representing stan code
 #' @return A `character` vector of the stan code without profiling statements
 remove_profiling <- function(s) {
-  while (grepl("profile\\(.+\\)\\{", s, perl = TRUE)) {
+  while (grepl("profile\\(.+\\) \\{", s, perl = TRUE)) {
     s <- gsub(
-      "profile\\(.+\\)\\{((?:[^{}]++|\\{(?1)\\})++)\\}", "\\1", s, perl = TRUE
+      "profile\\(.+\\) \\{((?:[^{}]++|\\{(?1)\\})++)\\}", "\\1", s, perl = TRUE
     )
   }
   return(s)


seabbs:
@adrian-lison I've tagged you to review as I feel like in general, it is good for all code contributions to be seen by two pairs of eyes. No pressure though, there isn't a great rush, and happy to self-review if not available.

R/model.R
```diff
@@ -229,8 +239,9 @@ enw_inits <- function(data) {
 #' @param verbose Logical, defaults to `TRUE`. Should verbose
 #' messages be shown.
 #'
-#' @param stanc_options A list of options to pass to the `stanc_options` of [cmdstanr::cmdstan_model()]
-#' by default "01" is passed which specifies simple optimisations should be done by the prior to compilation
+#' @param stanc_options A list of options to pass to the `stanc_options` of
+#' [cmdstanr::cmdstan_model()] by default "01" is passed which specifies simple
+#' optimisations should be done by the prior to compilation
```
adrian-lison, line 244: #' @param stanc_options A list of options to pass to the `stanc_options` of
#' [cmdstanr::cmdstan_model()]**. By** default "01" is passed, which specifies simple
#' optimisations should be done by **cdmstan** prior to compilation.

inst/stan/functions/obs_lpmf.stan
```diff
@@ -1,20 +1,32 @@
...
+    l = sl[i];
+    start_snap = csl[i] - start_n - l;
+    vector[l] exp_obs_by_rep;
+    vector[l] rdlh;
     // Find final observed/imputed expected observation
-    tar_obs = imp_obs[sg[i]][st[i]];
+    tar_obs = imp_obs[g][t];
     // allocate report day effects
-    rdlh = srdlh[rdlurd[st[i]:(st[i] + sl[i] - 1), sg[i]]];
+    rdlh = srdlh[rdlurd[t:(t + l - 1), g]];
```
adrian-lison, line 24: Like the new style with indices precomputed, makes it more readable!

inst/stan/functions/obs_lpmf.stan
```diff
@@ -1,20 +1,32 @@
...
     // combine expected final obs and date effects to get expected obs
-    exp_obs = expected_obs(tar_obs, ref_lh[1:sl[i], dpmfs[i]], rdlh, ref_p);
-    // observation error model
-    tar += neg_binomial_2_lupmf(obs[i, 1:sl[i]] | exp_obs, phi);
+    exp_obs_by_rep = expected_obs(tar_obs, ref_lh[1:l, dpmfs[i]], rdlh, ref_p);
+    exp_obs[(start_snap + 1):(start_snap + l)] = exp_obs_by_rep;
```
adrian-lison, line 27: This is really a question of taste, but I would also be fine with directly assigning to the slice of `exp_obs`, such that `exp_obs_by_rep` would not be needed.

inst/stan/functions/obs_lpmf.stan
```diff
@@ -1,20 +1,32 @@
-real obs_lpmf(int[] dummy, int start, int end, int[,] obs, int[] sl,
+real obs_lpmf(int[] dummy, int start, int end, int[] obs, int[] sl, int[] csl,
               vector[] imp_obs, int[] sg, int[] st, int[,] rdlurd,
               vector srdlh, matrix ref_lh, int[] dpmfs, int ref_p, real phi) {
   real tar = 0;
   real tar_obs;
-  int t;
+  int init_cl = sl[start];
```
adrian-lison, line 6: `init_cl` seems to be unused

inst/stan/epinowcast.stan
```diff
@@ -156,8 +159,8 @@ model {
   sqrt_phi ~ normal(sqrt_phi_p[1], sqrt_phi_p[2]) T[0,];
   // log density: observed vs model
   if (likelihood) {
-    target += reduce_sum(obs_lupmf, st, 1, obs, sl, imp_obs, sg, st, rdlurd,
-                         srdlh, ref_lh, dpmfs, ref_p, phi);
+    target += reduce_sum(obs_lupmf, st, 1, flat_obs, sl, csl, imp_obs, sg, st,
+                         rdlurd, srdlh, ref_lh, dpmfs, ref_p, phi);
```
adrian-lison, line 163: If I am not mistaken, doing it this way does not allow the scheduler to select slices with only parts of a reference date _x_ group observation (i.e. only parts of the corresponding row in the reporting triangle). I think this is fine (the big advantage of doing it this way is that we can have the selection from `sg`, `st` and `sl` directly using the indices between `start:end`). Just to keep in mind that we have a tiny bit of additional flexibility for the scheduler left in case we want to change this later.

inst/stan/functions/obs_lpmf.stan
```diff
@@ -1,20 +1,32 @@
...
+  vector[n] exp_obs;
+  int start_snap, g, t, l;
+
   for (i in start:end) {
-    vector[sl[i]] exp_obs;
-    vector[sl[i]] rdlh;
+    g = sg[i];
+    t = st[i];
+    l = sl[i];
+    start_snap = csl[i] - start_n - l;
```
adrian-lison, line 18: Just an idea regarding readibility: if we are using this loop anyway, wouldn't it be more intuitive code to initialize `start_snap` at 0* and then increment it by `l` in every iteration?

*or 1, as you prefer - would need to change slice of `exp_obs` below as well

adrian-lison: And rather call it `start_index` or so, to make clear that this is referencing the index of the `exp_obs` vector?

inst/stan/epinowcast.stan
```diff
@@ -8,16 +8,19 @@ functions {
...
   int s; // number of snapshots there are
   int g; // number of data groups
   int st[s]; // when in this time snapshots are from
   int ts[t, g]; // snapshot related  to time and group
   int sl[s]; // how many days of reported data does each snapshot have
+  int csl[s]; // cumulative version of the above
   int sg[s]; // how snapshots are related
   int dmax; // maximum possible report date
   // Observations
   int obs[s, dmax]; // obs for each primary date (row) and report date (column)
```
adrian-lison, line 22: We keep `obs` for now because it makes the generated quantities easier, right?

R/model.R
```diff
@@ -60,15 +60,22 @@ enw_obs_as_data_list <- function(pobs) {
   )
   latest_matrix <- as.matrix(latest_matrix[, -1])
 
+  # get new confirm for processing
+  new_confirm <- data.table::copy(pobs$new_confirm[[1]])
+  data.table::setorderv(new_confirm, c("reference_date", "group",  "delay"))
+
+  # get flat observations
+  flat_obs <- new_confirm$new_confirm
```
adrian-lison, line 68: Is there any risk regarding sorting order?

inst/stan/epinowcast.stan
```diff
@@ -8,16 +8,19 @@ functions {
...
 data {
   // Indexes and lookups
+  int n; // total observations
   int t; // time range over which data is available 
   int s; // number of snapshots there are
   int g; // number of data groups
   int st[s]; // when in this time snapshots are from
   int ts[t, g]; // snapshot related  to time and group
   int sl[s]; // how many days of reported data does each snapshot have
+  int csl[s]; // cumulative version of the above
```
adrian-lison, line 18: Out of interest, have you also thought about the alternative strategy of postponing the flattening to the model and doing it in `generated_data`? It would be slightly less efficient, because the flattening is done once by each chain. The advantage could be to have a more concise and intuitive data list on the other hand... I am just interested in your thoughts, quite impartial regarding this...

R/model.R
```diff
@@ -77,22 +84,25 @@ enw_obs_as_data_list <- function(pobs) {
...
   data <- list(
+    n = length(flat_obs),
     t = pobs$time[[1]],
     s = pobs$snapshots[[1]],
     g = pobs$groups[[1]],
     st = snap_time,
     ts = snap_lookup,
     sl = snap_length,
-    sg = unique(pobs$new_confirm[[1]][, .(reference_date, group)])$group,
+    csl = cumsum(snap_length),
```
adrian-lison, line 101: That's a clever way to approach this :)

adrian-lison (approved):
Well implemented, looks all correct to me. Just some questions and ideas, mainly regarding readibility of the code. I agree that the "biggest" efficiency improvement should come from a vectorization of the final likelihood statement. It would be nice to know which grain size was chosen by the scheduler though, because if the grainsize is very small, then the vectorization may not have a big effect anyway. I would expect this to be more relevant in case without within-chain parallelization. We can explore this further when doing the parallelization for the model version from #43 in a later PR I think.

inst/stan/epinowcast.stan
seabbs: So I did consider this but I have been trying to keep as much code in R vs in stan as possible so that it can be unit tested (obviously not at the moment) which was the main driver behind this approach. The other reason is that if we transition the back-end or go back-end agnostic in some far future having more code in R vs in stan is likely going to be helpful.

R/model.R
seabbs: So I put the sorting here as yes if the data is resorted after preprocessing but before any modelling there will be some issues. I think having sorting here though mitigates this risk though there could be some edge risk with the other datasets. Probably worth hardening at a later date.

inst/stan/functions/obs_lpmf.stan
seabbs: Good idea - will add this.

inst/stan/epinowcast.stan
seabbs: Yes, you are correct. Useful to flag.

inst/stan/functions/obs_lpmf.stan
seabbs: Yes, I think I agree here. Will update.

inst/stan/functions/obs_lpmf.stan
seabbs: :)

PR #45, opened 2022-04-12
Update CmdStan version

adrian-lison:
Since we added code optimization with stanc, we require CmdStan 2.29 at least.


seabbs (approved):
Good catch.

PR #46, opened 2022-04-12
Bugfix optional profiling

### LINKED ISSUE #44: Bug: When profile = TRUE generated quantities are not present
When `profile = TRUE` and profiling is in use as introduced in #10 then generated quantities are no longer stored. This means that nowcasts etc do not exist and plotting is not possible. As this feature is only used during development for a limited range of purposes this is not a must fix. Flagging @adrian-lison who implemented #10 and may have some insights. 

The problem can be seen by running the readme example code with `profile = TRUE` in `enw_model`.

adrian-lison:
This PR fixes #44. In addition, it makes the regex for the optional profiling white-space independent.

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -14,3 +14,5 @@
 ^vignettes/cache$
 ^vignettes/germany-age-stratified-nowcasting.Rmd.orig\.Rmd\.orig$
 ^vignettes/precompile.R$
+^.*\.Rproj$
+^\.Rproj\.user$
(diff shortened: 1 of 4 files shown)


adrian-lison:
PS.: Sorry for the non-linearity in the commit history. I wanted to use rebase, but it did not work somehow...

seabbs (approved):
Looks good to me. Good to know that can't profile variable declaration. Sensible to tighten up the regex (I really need to brush up on regexing it is not a strong point)

PR #48, opened 2022-04-13
Fix bug causing pkgdown to fail

### LINKED ISSUE #47: Bug: Pkgdown site failing to build
Currently the `pkgdown` site failing to build (see https://github.com/epiforecasts/epinowcast/runs/5998039307?check_suite_focus=true). This looks like it was introduced in #41 (tagging @adrian-lison). 

From a quick skim it looks like it is because the package index is now incomplete. This can be fixed by adding a concept (for example util) to the documentation of the new `remove_profiling` function.

seabbs:
This PR fixes #47 which caused pkgdown to fail. It adds a concept to `remove_profiling` and moves it into the `model-tools.R` file.


adrian-lison (approved):
Thanks for fixing this. I didn't know that concepts are also needed for internal functions, now I know :)

PR #51, opened 2022-04-14
Feature vectorised truncated dists

seabbs:
Should contain the vectorised truncation changes (#38) and the merged in changes from #42 prior to reverting.


adrian-lison (dismissed):
I approve!

PR #54, opened 2022-05-10
Feature optional profiling include paths

adrian-lison:
This PR extends the optional profiling also to .stan files in the include paths, and moves the whole functionality to a separate function which is called by `enw_model` if required. All manipulated .stan files with profiling statements removed are stored in the same temporary directory.

Implementation details: By using the same folder structure in the temporary directory as in the original include paths, the relative include paths still apply and the model code is not further touched aside from removing the profiling. This should make the approach very robust. Also, the names of the main model and the included file paths are now directly mirrored, so that warnings or errors from stanc stating a filepath allow the user to pinpoint the corresponding original .stan file.

This PR also adds more detailed profiling statements to the likelihood parts of the model.


inst/stan/functions/obs_lpmf.stan
```diff
@@ -17,18 +17,27 @@ real obs_lpmf(int[] dummy, int start, int end, int[] obs, int[] sl, int[] csl,
...
+    ref_lh_i = ref_lh[1:l, dpmfs[i]];
     // allocate report day effects
     rdlh = srdlh[rdlurd[t:(t + l - 1), g]];
+    }
     // combine expected final obs and date effects to get expected obs
+    profile("model_likelihood_expected_obs") {
     esnap += l;
     exp_obs[ssnap:esnap] = expected_obs(
-      tar_obs, ref_lh[1:l, dpmfs[i]], rdlh, ref_p
+      tar_obs, ref_lh_i, rdlh, ref_p
```
adrian-lison, line 33: The reference day effects allocation was moved out of the `expected_obs` call to be able to profile it independently.

seabbs (changes requested):
This looks really nice and as I send before worth considering for an upstream contribution to cmsdstanr. I have some very boring style points and would like the linting sorted out 😄

adrian-lison:
Thanks for reviewing @seabbs :) I have improved the formatting as suggested. Will consider proposing it to `cmdstanr` also :)

seabbs (approved):
LGTM to me. Just updating the docs and then will merge (and add news etc direct to master).

PR #60, opened 2022-06-14
Add inv_sqrt

seabbs:
This PR adds support for `inv_sqrt` (https://mc-stan.org/docs/2_18/functions-reference/power-and-logarithm-functions.html. No change in output or performance impact.


adrian-lison (approved):
Ready to roll!

PR #61, opened 2022-06-14
Bug: Fix proper scoring

### LINKED ISSUE #35: Update to new scoringutils interface
@nikosbosse has just released a new version of `scoringutils` which has an updated interface.  We need to update the optional function supplied in the package to make using `scoringutils` easier for package users to use this interface.

https://github.com/epiforecasts/epinowcast/blob/735c6c48ec1eb93edeb72ca06b7325493b44672b/R/model-validation.R#L19

`scoringutils` changelog: https://epiforecasts.io/scoringutils/news/index.html

seabbs:
This PR closes #35 by adding support for the new `scoringutils` interface. On top of this it adds examples for using this function and updates the vignette. 

It also  adds the basic example framework I have used elsewhere in compute heavy packages (https://github.com/epiforecasts/forecast.vocs) which can be further generalised, used to power examples for other functions, and as the basis for building out a testing suite (again `forecast.vocs` represents how I would go about doing this).


seabbs:
Merging this in as fixes live bug. Happy to update after review

Issue #52, opened 2022-04-14
More informative error for negative new confirmed cases


adrian-lison:
When the difference between cumulative confirmed cases for a certain reference date is negative, `enw_reporting_triangle` will throw an error. However, there could be two different ways to arrive at such a scenario:

- the reporting data indeed has "negative" reports for certain reference dates, for example due to corrections 
- the user has accidentally provided data in a wrong format, i.e. "confirmed" is actually "new confirmed" already

I am not sure how much we want to "shepherd" the user, but it could be useful to check for the latter case and throw a different error message than in the first case.

seabbs:
This is a good point. It would be great to get this in to the current development version as a nice way to safe guard users.

adrian-lison:
ToDos: Also consider discussion in #62, how do we synchronize between `new_confirm` and `prop_reported`. Set prop_reported to zero as well if we remove negative counts?

And decide whether we want to use a new variable (i.e `new_confirm_adj`) to adjust the data instead of the original variable to better inform the user.

PR #62, opened 2022-06-20
Feature: Add empirical reporting proportions and handle max confirmed

seabbs:
This PR improves handling of notifications outside of the maximum delay, makes this optional, and adds some quality of life features to make plotting + exploration of empirical reporting easier. PR also adds minimal preprocessing documentation and testing.

* Adds additional quality of life data processing so that the maximum number (`max_confirm`) of notifications is available in every row (for both cumulative and incidence notifications) and the cumulative and daily empirical proportion reported are calculated for the user during pre-processing. 

* The default approach to handling reported notifications beyond the maximum delay has been changed. In `0.0.6` and previous versions notifications beyond the maximum delay were silently dropped. In `0.0.7` this is now optional behaviour (set using `max_delay_strat` in [enw_preprocess_data()]) and the default is instead to add these notifications to the last included delay were present. This should produce more accurate long-term nowcasts when data is available but means that reported notifications for the maximum delay need to be interpreted with this in mind.


seabbs:
LGTM (self-review). Happy to revert if there are issues.

R/preprocess.R
```diff
@@ -106,6 +106,42 @@ enw_assign_group <- function(obs, by = c()) {
...
+#' @description FUNCTION_DESCRIPTION
+#' @param obs PARAM_DESCRIPTION
+#'
+#' @return OUTPUT_DESCRIPTION
+#'
+#' @family preprocess
+#' @export
+#' @importFrom data.table copy
+enw_add_max_reported <- function(obs) {
+  obs <- data.table::copy(obs)
```
adrian-lison, line 134: Do we have a clear strategy when to copy `obs`? In this case, e.g. `enw_latest_data` will also copy `obs` again, so it looks like a redundant copy to me.

R/preprocess.R
```diff
@@ -193,20 +258,61 @@ enw_new_reports <- function(obs) {
     by = c("group")
   ]
   reports <- reports[, delay := 0:(.N - 1), by = c("reference_date", "group")]
+
+  if (!is.null(reports$max_confirm)) {
+    reports[, prop_reported := new_confirm / max_confirm]
+  }
+
+  if (set_negatives_to_zero) {
+    reports <- reports[new_confirm < 0, new_confirm := 0]
```
adrian-lison, line 267: Will lead to inconsistencies between `new_confirm` and `prop_reported`. Could be better to have them synchronized, i.e. set `prop_reported` to zero as well..

Another thing to note is that max_confirm will currently be smaller than the sum of `new_confirm` if there are negative entries. Not sure if this is an issue, and what is more important: internal consistency in the dataframe vs. reminding the user that this data has been manipulated by having the original max_confirm remain as it was

R/preprocess.R
```diff
@@ -193,20 +258,61 @@ enw_new_reports <- function(obs) {
...
+#' @inheritParams enw_preprocess_data
+#' @family preprocess
+#' @export
+#' @importFrom data.table copy
+#' @examples
+#' obs <- enw_example("preprocessed")$obs[[1]]
+#' enw_filter_obs(obs, max_delay = 2)
+enw_filter_obs <- function(obs, max_delay) {
+  obs <- data.table::copy(obs)
+  obs <- obs[, .SD[report_date <= (reference_date + max_delay - 1)],
```
adrian-lison, line 287: This excludes the maximum delay, unlike stated in the function `@return` description.

R/preprocess.R
```diff
@@ -233,51 +345,173 @@ enw_reporting_triangle_to_long <- function(obs) {
...
+enw_construct_data <- function(obs, new_confirm, latest, reporting_triangle,
+                               metareport, metareference, max_delay) {
+  out <- data.table::data.table(
+    obs = list(obs),
+    new_confirm = list(new_confirm),
+    latest = list(latest),
+    reporting_triangle = list(reporting_triangle),
+    metareference = list(metareference),
+    metareport = list(metareport),
+    time = nrow(latest[group == 1]),
```
adrian-lison, line 394: I think this only works if all groups have the same reporting dates in the `obs` provided. But this is addressed in #64.

R/preprocess.R
```diff
@@ -233,51 +345,173 @@ enw_reporting_triangle_to_long <- function(obs) {
...
+    choices = c("exclude", "add_to_max_delay")
+  )
   obs <- data.table::as.data.table(obs)
   obs <- obs[order(reference_date)]
 
-  if (!missing(min_report_date)) {
-    obs <- obs[report_date >= min_report_date]
-  }
-
-  # assign groups
```
adrian-lison, line 267: We are dropping the `min_report_date` option?

R/preprocess.R
```diff
@@ -233,51 +345,173 @@ enw_reporting_triangle_to_long <- function(obs) {
...
-  obs <- obs[, .SD[report_date <= (reference_date + max_delay - 1)],
-    by = c("reference_date", "group")
-  ]
+  if (max_delay_strat %in% "add_to_max_delay") {
+    obs[
+      report_date == (reference_date + max_delay - 1),
+      confirm := max_confirm,
+      by = "group"
+    ]
+  }
```
adrian-lison, line 508: As mentioned earlier, I am still confused by the current definition of `max_delay` here. As I understand, we are still excluding the maximum delay and only modeling until the day _before_ the maximum delay. Or am I getting something wrong?

R/preprocess.R
```diff
@@ -297,38 +531,28 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
...
-    max_delay = max_delay,
-    max_date = max(obs$report_date)
+  out <- enw_construct_data(
+    obs = obs,
+    new_confirm = diff_obs,
+    latest = latest,
+    reporting_triangle = reporting_triangle,
+    metareference = metareference,
+    metareport = metareport,
+    max_delay = max_delay
```
adrian-lison, line 554: Good idea!

adrian-lison (commented):
Nice work, added a couple of comments :)

tests/testthat/test-enw_preprocess_data.R
```diff
@@ -0,0 +1,52 @@
...
+# Filter example hospitalisation data to be natioanl and over all ages
+nat_germany_hosp <- germany_covid19_hosp[location == "DE"]
+nat_germany_hosp <- nat_germany_hosp[age_group %in% "00+"]
+
+cols <- c(
+  "obs", "new_confirm", "latest", "reporting_triangle",
+  "metareference", "metareport", "time", "snapshots", "groups",
+  "max_delay", "max_date"
+)
+test_that("Preprocessing produces expected output with default settings", {
```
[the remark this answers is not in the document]
seabbs: yes exactly but we should add these for each function (i.e enw_new_confirm) etc vs the whole of the preprocessing chain (though we may need something better here as well)

R/preprocess.R
seabbs: That is a good point. I think in general we don't ever want to depend on the input being a `data.table` and we also don't want to ever alter in place someone's input. For internal functions though we can do both those things. At the moment gone pretty maximal with exported functions. We can definitely pair those down as we stabilise the package. 

In terms of this copy you are right and this looks like a duplicate 😢

R/preprocess.R
seabbs: Yeah, this is a good point and as in #52 we need to be more nuanced with what we are doing here. 

So there is a tension between cleaning the data as we need and highlighting we have done so to the user. Perhaps we should be making a new variable (i.e `new_confirm_adj`) and using that to adjust the data etc? 

Otherwise I agree moving setting negatives to zero above calculating the proportion reported makes sense. Also agree that leaving max_confirm as is makes sense to flag to the user.

R/preprocess.R
seabbs: yes, it does and yup. It is a bit of a shame we have this requirement but at least for now it makes things easier.

R/preprocess.R
seabbs: So I thought it should be the maximum delay assuming it is defined from 0 which is what we have in the stan code etc. That being said very happy to change that definition.

PR #64, opened 2022-06-24
Feature missing reference dates

adrian-lison:
This is a WIP draft pull request for the missing reference date feature. It implements only the absolute minimum to get a running model, i.e. the preprocessing and the minimal running version of the stan model. No postprocessing yet. I have tested the model with simulated data (see vignette draft in extra dev branch "missing_reference_vignette_wip") and it seems to be working fine. I will add a couple of comments to the coding decisions and add remarks on currently missing extensions that are planned for the MVP.

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -19,7 +19,9 @@ enw_priors <- function() {
       "logmean_sd",
       "logsd_sd",
       "rd_eff_sd",
-      "sqrt_phi"
+      "sqrt_phi",
+      "alpha_start",
+      "alpha_sd"
     ),
     description = c(
       "Standard deviation for expected final observations",
@@ -28,7 +30,9 @@ enw_priors <- function() {
       "Standard deviation of scaled pooled logmean effects",
       "Standard deviation of scaled pooled logsd effects",
       "Standard deviation of scaled pooled report date effects",
-      "One over the square of the reporting overdispersion"
+      "One over the square of the reporting overdispersion",
+      "Logit start value for share of cases with known reference date",
+      "Standard deviation of random walk for share of cases with known reference date"
     ),
     distribution = c(
       "Zero truncated normal",
@@ -37,10 +41,12 @@ enw_priors <- function() {
       "Zero truncated normal",
       "Zero truncated normal",
       "Zero truncated normal",
+      "Zero truncated normal",
+      "Normal",
       "Zero truncated normal"
     ),
-    mean = c(0, 1, 0.5, rep(0, 4)),
-    sd = rep(1, 7)
+    mean = c(0, 1, 0.5, rep(0, 4), 0, 0),
+    sd = c(rep(1, 7), 1, 0.1)
   )
 }
 
--- a/R/model.R
+++ b/R/model.R
@@ -63,6 +63,8 @@ enw_obs_as_data_list <- function(pobs) {
   # get new confirm for processing
   new_confirm <- data.table::copy(pobs$new_confirm[[1]])
   data.table::setorderv(new_confirm, c("reference_date", "group", "delay"))
+  new_confirm_missing <- data.table::copy(pobs$new_confirm_missing[[1]])
+  data.table::setorderv(new_confirm_missing, c("group", "reference_date"))
 
   # get flat observations
   flat_obs <- new_confirm$new_confirm
@@ -87,6 +89,13 @@ enw_obs_as_data_list <- function(pobs) {
   snap_time <- unique(new_confirm[, .(reference_date, group)])
   snap_time[, t := 1:.N, by = "group"]
   snap_time <- snap_time$t
+  
+  # obs with missing reference date
+  obs_miss <- as.matrix(data.table::dcast(
+    new_confirm_missing,
+    group ~ report_date,
+    value.var = "new_confirm", fill = 0
+  )[, -1])
 
   # Format indexing and observed data
   # See stan code for docs on what all of these are
@@ -103,7 +112,8 @@ enw_obs_as_data_list <- function(pobs) {
     dmax = pobs$max_delay[[1]],
     obs = as.matrix(pobs$reporting_triangle[[1]][, -c(1:2)]),
     flat_obs = flat_obs,
-    latest_obs = latest_matrix
+    latest_obs = latest_matrix,
+    obs_miss = obs_miss
   )
   return(data)
 }
--- a/R/preprocess.R
+++ b/R/preprocess.R
@@ -138,9 +138,9 @@ enw_retrospective_data <- function(obs, rep_date, rep_days, ref_date,
   retro_data <- retro_data[report_date <= rep_date]
 
   if (!missing(ref_days)) {
-    ref_date <- max(retro_data$reference_date) - ref_days
+    ref_date <- max(retro_data$reference_date, na.rm = TRUE) - ref_days
   }
-  retro_data <- retro_data[reference_date >= ref_date]
+  retro_data <- retro_data[reference_date >= ref_date | is.na(reference_date)]
   return(retro_data[])
 }
 
--- a/R/preprocess.R
+++ b/R/preprocess.R
@@ -238,6 +238,56 @@ enw_reporting_triangle_to_long <- function(obs) {
 
 #' @title FUNCTION_TITLE
 #' @description FUNCTION_DESCRIPTION
+#' @param obs A `data.table` with the observations to complete
+#' 
+#' @param min_date The first date from which to complete.
+#' 
+#' @param max_date The last date until which to complete.
+#' 
+#' @param groups A `data.table` with one row per group, a column for each group
+#' attribute and a column `group` holding the group ID.
+#' 
+#' @param max_delay The maximum reporting delay.
+#' 
+#' @param include_missing Should entries for cases with missing reference date
+#' be completed as well?, Default: FALSE
+#' 
+#' @return A `data.table` with completed entries for all combinations of reference
+#' dates, groups and possible report dates.
+#' @family preprocess
+enw_complete_dates <- function(obs, min_date, max_date, groups, max_delay, include_missing = FALSE) {
+  obs <- data.table::as.data.table(obs)
+  dates = seq.Date(min_date,max_date,by=1)
+  completion <- CJ(
+    reference_date = dates,
+    group = groups$group,
+    report_date = 0:(max_delay - 1)
+  )
+  completion <- completion[, report_date := reference_date + report_date]
+  completion <- completion[report_date <= max_date]
+
+  if (include_missing) {
+    completion <- rbind(
+      completion,
+      CJ(
+        reference_date = as.Date(NA),
+        group = groups$group,
+        report_date = dates
+      )
+    )
+  }
+  # join completion with groups and original obs
+  completion <- completion[groups, on = "group"]
+  obs <- obs[completion, on = c("reference_date", "report_date", names(groups))]
+  # impute
+  obs[, confirm := nafill(nafill(confirm, "locf"), fill = 0),
+    by = c("reference_date", "group")
+  ]
+  return(obs)
+}
+
+#' @title FUNCTION_TITLE
+#' @description FUNCTION_DESCRIPTION
 #' @param obs PARAM_DESCRIPTION
 #'
 #' @param by PARAM_DESCRIPTION, Default: c()
@@ -269,17 +319,17 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
 
   # assign groups
   obs <- enw_assign_group(obs, by = by)
+  by_with_group_id <- c("group",by)
   
   # complete missing report dates
-  comp <- rbind(
-    obs[!is.na(reference_date)][rep(1:.N,each=max_delay), .(report_date = reference_date + 0:(.N - 1)), by = c("reference_date", "group")],
-    CJ(reference_date = as.Date(NA), group = unique(obs[,group]), report_date = obs[,seq.Date(min(reference_date,na.rm=T),
-                                                                                              max(reference_date,na.rm=T)+max_delay-1,by=1)])
-  )
-  grouping_factors <- obs[, lapply(.SD, min, na.rm = TRUE), .SDcols = by, by = "group"]
-  comp <- comp[grouping_factors, on = "group"]
-  obs <- obs[comp, on = c("reference_date", "group", "report_date", by)]
-  obs[, confirm:=nafill(nafill(confirm, "locf"), fill = 0), by = c("reference_date", "group")]
+  # Note: it might be better to get the "now" date as last date instead of
+  # estimating it from data using the max report date
+  obs <- enw_complete_dates(obs,
+                            min_date = min(obs$reference_date,na.rm=T),
+                            max_date =  max(obs$report_date),
+                            groups = unique(obs[,..by_with_group_id]),
+                            max_delay = max_delay,
+                            include_missing = any(is.na(obs$reference_date)))
   
   # filter by maximum reporting delay
   obs <- obs[, .SD[report_date <= (reference_date + max_delay - 1) | is.na(reference_date)],
--- a/R/preprocess.R
+++ b/R/preprocess.R
@@ -269,7 +269,18 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
 
   # assign groups
   obs <- enw_assign_group(obs, by = by)
-
+  
+  # complete missing report dates
+  comp <- rbind(
+    obs[!is.na(reference_date)][rep(1:.N,each=max_delay), .(report_date = reference_date + 0:(.N - 1)), by = c("reference_date", "group")],
+    CJ(reference_date = as.Date(NA), group = unique(obs[,group]), report_date = obs[,seq.Date(pmin(min(report_date,na.rm=T),min(reference_date,na.rm=T)),
+                                                                                              pmin(max(report_date,na.rm=T),max(reference_date,na.rm=T)),by=1)])
+  )
+  grouping_factors <- obs[, lapply(.SD, min, na.rm = TRUE), .SDcols = by, by = "group"]
+  comp <- comp[grouping_factors, on = "group"]
+  obs <- obs[comp, on = c("reference_date", "group", "report_date", by)]
+  obs[, confirm:=nafill(nafill(confirm, "locf"), fill = 0), by = c("reference_date", "group")]
+  
   # filter by maximum report date
   obs <- obs[, .SD[report_date <= (reference_date + max_delay - 1) | is.na(reference_date)],
     by = c("reference_date", "group")
--- a/R/preprocess.R
+++ b/R/preprocess.R
@@ -273,8 +273,8 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
   # complete missing report dates
   comp <- rbind(
     obs[!is.na(reference_date)][rep(1:.N,each=max_delay), .(report_date = reference_date + 0:(.N - 1)), by = c("reference_date", "group")],
-    CJ(reference_date = as.Date(NA), group = unique(obs[,group]), report_date = obs[,seq.Date(pmin(min(report_date,na.rm=T),min(reference_date,na.rm=T)),
-                                                                                              pmin(max(report_date,na.rm=T),max(reference_date,na.rm=T)),by=1)])
+    CJ(reference_date = as.Date(NA), group = unique(obs[,group]), report_date = obs[,seq.Date(min(reference_date,na.rm=T),
+                                                                                              max(reference_date,na.rm=T)+max_delay-1,by=1)])
   )
   grouping_factors <- obs[, lapply(.SD, min, na.rm = TRUE), .SDcols = by, by = "group"]
   comp <- comp[grouping_factors, on = "group"]
--- a/R/preprocess.R
+++ b/R/preprocess.R
@@ -192,7 +192,7 @@ enw_new_reports <- function(obs) {
   reports[, new_confirm := confirm - data.table::shift(confirm, fill = 0),
     by = c("reference_date", "group")
   ]
-  reports <- reports[, .SD[reference_date >= min(report_date)],
+  reports <- reports[, .SD[reference_date >= min(report_date) | is.na(reference_date)],
     by = c("group")
   ]
   reports <- reports[, delay := 0:(.N - 1), by = c("reference_date", "group")]
@@ -271,7 +271,7 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
   obs <- enw_assign_group(obs, by = by)
 
   # filter by maximum report date
-  obs <- obs[, .SD[report_date <= (reference_date + max_delay - 1)],
+  obs <- obs[, .SD[report_date <= (reference_date + max_delay - 1) | is.na(reference_date)],
     by = c("reference_date", "group")
   ]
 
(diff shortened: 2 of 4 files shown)


inst/stan/epinowcast.stan
```diff
@@ -39,6 +35,11 @@ data {
   matrix[urds, nrd_effs + 1] rd_fixed; // design matrix for report dates
   int nrd_eff_sds; // number of standard deviations to use for pooling for rds
   matrix[nrd_effs, nrd_eff_sds + 1] rd_random; // Pooling pmf design matrix 
+  // Observations
+  int obs[s, dmax]; // obs for each primary date (row) and report date (column)
+  int flat_obs[n]; // obs stored as a flat vector
+  int obs_miss[g, t]; // obs with missing primary date (group first)
```
adrian-lison, line 41: I put group first for reasons of memory locality. The time is used in a vectorized statement in the likelihood and is therefore the fast index, the groups are outer looped over and are therefore the slow index. Not sure if we want to change latest_obs as well for for consistency: the current order for latent obs is actually the more efficient one because there group is the fast index, on the other hand it is only used in the transformed data anyway.

inst/stan/epinowcast.stan
```diff
@@ -76,6 +79,9 @@ parameters {
   vector<lower=0>[neff_sds] logmean_sd; // pooled modifiers to logmean
   vector<lower=0>[neff_sds] logsd_sd; // pooled modifiers to logsd
   vector<lower=0>[nrd_eff_sds] rd_eff_sd; // pooled modifiers to report date
+  real alpha_start[g]; // starting value for alpha
+  real<lower=0> alpha_sd; // standard deviation of the random walk increments
+  vector<offset=0, multiplier=alpha_sd>[t] alpha_epsilon[g]; // random walk increments, non-centered
```
adrian-lison, line 84: I am using the offset-multiplier style to declare non-centered parameterizations here...

inst/stan/epinowcast.stan
```diff
@@ -76,6 +79,9 @@ parameters {
   vector<lower=0>[neff_sds] logmean_sd; // pooled modifiers to logmean
   vector<lower=0>[neff_sds] logsd_sd; // pooled modifiers to logsd
   vector<lower=0>[nrd_eff_sds] rd_eff_sd; // pooled modifiers to report date
+  real alpha_start[g]; // starting value for alpha
+  real<lower=0> alpha_sd; // standard deviation of the random walk increments
```
adrian-lison, line 83: This implementation assumes the same scale of the random walk noise for all groups. Should we change this to be individual for each group? If yes, fixed or random effect? I know this may be taken over by a formula interface in the long term...

inst/stan/epinowcast.stan
```diff
@@ -128,6 +135,11 @@ transformed parameters{
     imp_obs[k] = exp(imp_obs[k]);
   }
   }
+  // estimate share of cases with eventually known reference date, modeled as
+  // a first order random walk for each group on the logit scale
+  for (k in 1:g) {
+    alpha[k] = inv_logit(alpha_start[k] + cumulative_sum(alpha_epsilon[k]));
```
adrian-lison, line 141: I am using my vectorized random walk version here. May be abandoned later if we want to use more complex models for `alpha` to.

inst/stan/epinowcast.stan
```diff
@@ -163,48 +175,92 @@ model {
...
   // reporting overdispersion (1/sqrt)
   sqrt_phi ~ normal(sqrt_phi_p[1], sqrt_phi_p[2]) T[0,];
   }
   // log density: observed vs model
   if (likelihood) {
     profile("model_likelihood") {
-    target += reduce_sum(obs_lupmf, st, 1, flat_obs, sl, csl, imp_obs, sg, st,
-                         rdlurd, srdlh, ref_lh, dpmfs, ref_p, phi);
+    target += obs_lupmf(flat_obs | obs_miss, dmax, sl, csl, g, imp_obs, sg, st,
+                         rdlurd, srdlh, ref_lh, dpmfs, ref_p, alpha, phi);
```
adrian-lison, line 191: Not sure if there is any logic about the order of arguments that I might have violated

inst/stan/epinowcast.stan
```diff
@@ -39,6 +35,11 @@ data {
   matrix[urds, nrd_effs + 1] rd_fixed; // design matrix for report dates
   int nrd_eff_sds; // number of standard deviations to use for pooling for rds
   matrix[nrd_effs, nrd_eff_sds + 1] rd_random; // Pooling pmf design matrix 
+  // Observations
```
adrian-lison, line 38: Just a general remark: the missingness model is not implemented as optional yet, I would add that in the end.

inst/stan/functions/obs_lpmf.stan
```diff
@@ -1,18 +1,19 @@
...
+  int n_obs = num_elements(obs);
+  int n_obs_miss = num_elements(obs_miss[1]);
+  vector[n_obs] exp_obs;
+  vector[n_obs_miss] exp_obs_miss[n_groups] = rep_array(rep_vector(0, n_obs_miss), n_groups);
   int g, t, l;
   int ssnap = 1;
   int esnap = 0;
 
-  for (i in start:end) {
+  for (i in 1:n_snaps) {
```
adrian-lison, line 16: The reduce sum functionality has been removed and I simplified some elements (e.g. start_n and end_n turned out to be senseless.

inst/stan/functions/obs_lpmf.stan
```diff
@@ -21,14 +22,25 @@ real obs_lpmf(int[] dummy, int start, int end, int[] obs, int[] sl, int[] csl,
...
     esnap += l;
-    exp_obs[ssnap:esnap] = expected_obs(
+    vector[l] exp_obs_all = expected_obs(
       tar_obs, ref_lh[1:l, dpmfs[i]], rdlh, ref_p
     );
+    // compute expected final obs with known and missing reference date
+    exp_obs[ssnap:esnap] = exp_obs_all * tar_alpha;
+    if(t+l-1>=1+dmax){
+      exp_obs_miss[g][max(1 + dmax, t):(t + l - 1)] += exp_obs_all[max(2 + dmax - t, 1):l] * (1 - tar_alpha);
+    }
```
adrian-lison, line 36: Here we have a potentially serious conflict with the current model setup. The problem is that we currently do not model the `exp_obs` for the maximum delay. While this decision is fine for the cases with known reference date, it is a problem for the cases with unknown reference date, because the exp_obs at the maximum delay are missing in the convolution over reference dates. This should bias the model, although not super strongly if the maximum delay is large enough. But it is a general issue to discuss. For now, I have just implemented the biased version.

inst/stan/functions/obs_lpmf.stan
```diff
@@ -21,14 +22,25 @@ real obs_lpmf(int[] dummy, int start, int end, int[] obs, int[] sl, int[] csl,
...
     ssnap += l;
   }
-  // observation error model (across all reference times and groups)
-  tar = neg_binomial_2_lupmf(snap_obs | exp_obs, phi);
+  // observation error model with known reference dates (across all reference times and groups)
+  tar = neg_binomial_2_lupmf(obs | exp_obs, phi);
+  // observation error model with missing reference dates
+  for (k in 1:n_groups) {
+    tar += neg_binomial_2_lupmf(obs_miss[k][(1 + dmax):n_obs_miss] | exp_obs_miss[k][(1 + dmax):n_obs_miss], phi);
+  }
```
adrian-lison, line 44: As can be seen, I did not go for a flattening of the `obs_miss` for now. Should discuss if we want this here as well.

inst/stan/functions/obs_lpmf.stan
```diff
@@ -21,14 +22,25 @@ real obs_lpmf(int[] dummy, int start, int end, int[] obs, int[] sl, int[] csl,
...
+    }
     ssnap += l;
   }
-  // observation error model (across all reference times and groups)
-  tar = neg_binomial_2_lupmf(snap_obs | exp_obs, phi);
+  // observation error model with known reference dates (across all reference times and groups)
+  tar = neg_binomial_2_lupmf(obs | exp_obs, phi);
+  // observation error model with missing reference dates
+  for (k in 1:n_groups) {
+    tar += neg_binomial_2_lupmf(obs_miss[k][(1 + dmax):n_obs_miss] | exp_obs_miss[k][(1 + dmax):n_obs_miss], phi);
```
adrian-lison, line 43: Currently, I am using exactly the same overdispersion as for the cases with known reference date. The question is whether we can think of typical scenarios in which the missingness has implications for the overdispersion. Then I should add a separate phi parameter for the `obs_miss`.

inst/stan/epinowcast.stan
```diff
@@ -163,48 +175,92 @@ model {
...
     profile("model_likelihood") {
-    target += reduce_sum(obs_lupmf, st, 1, flat_obs, sl, csl, imp_obs, sg, st,
-                         rdlurd, srdlh, ref_lh, dpmfs, ref_p, phi);
+    target += obs_lupmf(flat_obs | obs_miss, dmax, sl, csl, g, imp_obs, sg, st,
+                         rdlurd, srdlh, ref_lh, dpmfs, ref_p, alpha, phi);
     }
   }
 }
 
 generated quantities {
```
adrian-lison, line 196: The generated quantities are a bit of a mess / rather inefficient still and I clearly haven't helped things by adding the missing reference date stuff...

inst/stan/epinowcast.stan
```diff
@@ -163,48 +175,92 @@ model {
...
   int pp_obs[pp ? sum(sl) : 0];
+  int pp_obs_miss[pp ? sum(sl) : 0];
   vector[ologlik ? s : 0] log_lik;
+  vector[ologlik ? (t-dmax) : 0] log_lik_miss;
   int pp_inf_obs[cast ? dmax : 0, cast ? g : 0];
+  int pp_inf_obs_miss[cast ? t : 0, cast ? g : 0];
+  int pp_inf_obs_miss_rep[cast ? (t-dmax) : 0, cast ? g : 0];
```
adrian-lison, line 203: As discussed, we here restrict the reporting date - related quantities to the days after the maximum delay (starting at dmax + 1).

adrian-lison, line 203: There are to pp's for the missing reference date cases. One by reference date, one by reporting date. The one by reporting date could theoretically also be constructed during post-processing, not sure what is better.

inst/stan/epinowcast.stan
```diff
@@ -163,48 +175,92 @@ model {
...
+      pp_obs_tmp[i, 1:dmax] = neg_binomial_2_rng(exp_obs * tar_alpha, phi);
+      // realized observations with unknown reference date
+      pp_obs_tmp_miss[i, 1:dmax] = neg_binomial_2_rng(exp_obs * (1 - tar_alpha), phi);
       }
       profile("generated_loglik") {
       if (ologlik) {
+        // predicted observations with missing reference date by reporting date
+        if (i_time > 1) {
+          exp_obs_miss_rep[i_group][i_time:min(i_time + dmax - 1, t)]
+            += (exp_obs * (1 - tar_alpha))[1:min(dmax, t - i_time + 1)];
```
adrian-lison, line 238: Is this the way to line-break? ^^

inst/stan/epinowcast.stan
```diff
@@ -213,6 +269,18 @@ generated quantities {
...
+        int snap = ts[i, k];
+        pp_inf_obs_miss[i, k] = sum(pp_obs_tmp_miss[snap, 1:dmax]);
+      }
+      // cases with missing reference date (by reporting date)
+      for (i in 2:t) {
+        int snap = ts[i, k];
+        for (l in max(dmax - i + 2, 1):min(dmax, t - i + 1)){
+          pp_inf_obs_miss_rep[(i - dmax) + l - 1, k] += pp_obs_tmp_miss[snap, l];
+        }
+      }
```
adrian-lison, line 283: Again, we currently cannot include the maximum delay cases and are thus slightly underpredicting the `obs_miss`.

R/model.R
```diff
@@ -37,10 +41,12 @@ enw_priors <- function() {
...
       "Zero truncated normal",
       "Zero truncated normal",
+      "Zero truncated normal",
+      "Normal",
       "Zero truncated normal"
     ),
-    mean = c(0, 1, 0.5, rep(0, 4)),
-    sd = rep(1, 7)
+    mean = c(0, 1, 0.5, rep(0, 4), 0, 0),
+    sd = c(rep(1, 7), 1, 0.1)
```
adrian-lison, line 49: I was not 100% sure what sensible default priors to put here. Using an sd of 1 for the random walk seemed way to wide for me (would include up to a 7-fold increase in odds ratio in one time step in its 96% interval)... The current starting value prior is roughly between 10% and 90%.

R/preprocess.R
```diff
@@ -34,7 +34,7 @@ enw_add_metaobs_features <- function(metaobs, holidays = c(),
   metaobs <- data.table::copy(metaobs)
   metaobs[, day_of_week := weekdays(date)]
 
-  # make holidays be sundays
+  # make holidays be Sundays
```
adrian-lison, line 37: Rstudio was complaining, I am not a that much of a grammar nazi just by myself :D

R/preprocess.R
```diff
@@ -94,6 +94,9 @@ enw_extend_date <- function(metaobs, max_delay = 20) {
 #' @export
 #' @importFrom data.table as.data.table copy
 enw_assign_group <- function(obs, by = c()) {
+  if ("group" %in% names(obs)) {
+    stop("Dataset cannot have a column called 'group'.")
+  }
```
adrian-lison, line 99: I just came across this potential bug if the dataset has a column `group`. Should we refactor to using a less frequent name?

R/preprocess.R
```diff
@@ -266,13 +319,24 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
...
+  
+  # complete missing report dates
+  # Note: it might be better to get the "now" date as last date instead of
+  # estimating it from data using the max report date
+  obs <- enw_complete_dates(obs,
+                            min_date = min(obs$reference_date,na.rm=T),
+                            max_date =  max(obs$report_date),
+                            groups = unique(obs[,..by_with_group_id]),
+                            max_delay = max_delay,
+                            include_missing = any(is.na(obs$reference_date)))
```
adrian-lison, line 332: See comment. Should discuss if we want to have something like this. If the dataset does not go until the last report date, we could have a different time window than intended by the user.

R/preprocess.R
```diff
@@ -296,30 +360,37 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
   )
   diff_obs[, group := new_group][, new_group := NULL]
   obs[, old_group := NULL]
+  
+  # separate obs with and without missing reference date
+  reporting_available <- diff_obs[!is.na(reference_date)]
+  reporting_missing <- diff_obs[is.na(reference_date)]
```
adrian-lison, line 366: We could still run into trouble if the data provided by the user starts at a cumulative count that is nonzero. Should we create another issue for this?

R/preprocess.R
```diff
@@ -296,30 +360,37 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
...
     obs = list(obs),
-    new_confirm = list(diff_obs),
+    new_confirm = list(reporting_available),
+    new_confirm_missing = list(reporting_missing),
     latest = list(latest),
-    diff = list(diff_obs),
+    diff = list(reporting_available),
     reporting_triangle = list(reporting_triangle),
     metareference = list(metareference),
     metareport = list(metareport),
```
adrian-lison, line 396: I think the `time = nrow(latest[group == 1])` is only safe now after the completion of missing reporting delays, because now all groups should have the same number of observations.

inst/stan/epinowcast.stan
adrian-lison, line 238: In many places (e.g. here) I am assuming that `dmax<=t`. Otherwise, the model would break. Is this a reasonable assumption/requirement? And if yes, should we add a check for this somewhere?

inst/stan/epinowcast.stan
```diff
@@ -39,6 +35,11 @@ data {
   matrix[urds, nrd_effs + 1] rd_fixed; // design matrix for report dates
   int nrd_eff_sds; // number of standard deviations to use for pooling for rds
   matrix[nrd_effs, nrd_eff_sds + 1] rd_random; // Pooling pmf design matrix 
+  // Observations
+  int obs[s, dmax]; // obs for each primary date (row) and report date (column)
+  int flat_obs[n]; // obs stored as a flat vector
+  int obs_miss[g, t]; // obs with missing primary date (group first)
+  int latest_obs[t, g]; // latest obs for each snapshot group
```
adrian-lison, line 42: A currently unresolved issue with the latest obs is that the cases with missing reference dates are not included, slightly biasing the seeding. We have to decide on a sufficiently accurate, not to complicated way to include them. One way would be to do a rough back-projection (either assuming some backward reporting delay distribution or taking the empirical one). Should be discussed.

R/preprocess.R
```diff
@@ -233,6 +236,56 @@ enw_reporting_triangle_to_long <- function(obs) {
...
+#' 
+#' @param max_delay The maximum reporting delay.
+#' 
+#' @param include_missing Should entries for cases with missing reference date
+#' be completed as well?, Default: FALSE
+#' 
+#' @return A `data.table` with completed entries for all combinations of reference
+#' dates, groups and possible report dates.
+#' @family preprocess
+enw_complete_dates <- function(obs, min_date, max_date, groups, max_delay, include_missing = FALSE) {
```
adrian-lison, line 258: Just by the way: I believe the completion of the missing reporting dates can also be done using `dcast` in the case of the **reporting triangle** (and is actually already done). In other words, it is not necessary for the reporting triangle, but I think it is necessary for the other data structures. One option to think about would be to implement a more efficient version of the below function using dcast. Then again, this is more explicit and needs no reshaping of data structures...

inst/stan/epinowcast.stan
seabbs: Let's leave as the simple case for now as hopefully we will be ready to update to the formula interface in the short to medium term.

inst/stan/epinowcast.stan
seabbs: yup this is the better implementation. For now, could also update the expectation model one as well.

inst/stan/functions/obs_lpmf.stan
seabbs: We probably want to make this optional before going into a release candidate but good for now.

R/preprocess.R
```diff
@@ -296,30 +360,37 @@ enw_preprocess_data <- function(obs, by = c(), max_delay = 20,
...
-  # calculate reporting matrix
-  reporting_triangle <- enw_reporting_triangle(diff_obs)
+  # calculate reporting matrix on obs with available reference date
+  reporting_triangle <- enw_reporting_triangle(reporting_available)
 
   # extract latest data
-  latest <- enw_latest_data(obs)
+  # Note: currently, only the obs with available reference date are used
+  # This should to be extended to missing reference dates to avoid bias
+  latest <- enw_latest_data(reporting_available)
```
sbfnk, line 374: As implemented at the moment the filtering for reference dates in `enw_latest_data` will also need `na.rm = TRUE` I think.

inst/stan/functions/obs_lpmf.stan
sbfnk: At the moment there are no checks anywhere that `dmax` is not greater than the number of observations - so as implemented this is quite risky for going beyond the vector range of `obs_miss`.

seabbs:
> As implemented at the moment the filtering for reference dates in `enw_latest_data` will also need `na.rm = TRUE` I think (though not an issue here because no `ref_window` is given. 

I've resolved this in #82

R/preprocess.R
seabbs: Yes agree.

R/preprocess.R
seabbs: yes I think so.

R/preprocess.R
seabbs: Yup agree this is an issue. See [#81](https://github.com/epiforecasts/epinowcast/issues/81) for some concerns from @sbfnk about making this be to hardcoded to work on days. 

It is hard to know what the optimal course to take here is. Shall we have this as a step for now and look at changing in the future (though this will break old code) or have this as a function users can use themselves before beginning preprocessing?

R/preprocess.R
seabbs: yup that was intentional as happens quite a bit in data I've seen but I missed the fact it wouldn't enter into the metadata and so won't be assigned features/will break (I think). 

This looks quite efficient whilst also being clear so I'd go with this implementation vs `dcast` magic.

Doing it all at once (and maybe before preprocessing makes lots of sense).

R/preprocess.R
seabbs: yes agree. Shall we make a new issue. Would be good to get this into `0.1.0` as another breaking change.

R/model.R
seabbs: Yeah. I think this is fine for now. In general, going to need to think more about priors as it would be good to be able to pass effect size priors. I've done some work on this in my in progress `epict` package that I think can be ported. Not for this release though. Before resolving this someone (me!) should open an issue to do this.

inst/stan/epinowcast.stan
seabbs: 🙈

inst/stan/functions/obs_lpmf.stan
seabbs: > dmax is not greater than the number of observations

This is a good point. Is this something to enforce in the R code or to allow and then build resilience to here (i.e with a min).

> , I am using exactly the same overdispersion as for the cases with known reference date

I can see the argument either way. For simplicity at least for now I vote we go with a single overdispersion parameter and circle back to this as and when.

inst/stan/functions/obs_lpmf.stan
seabbs: 😡 😆

inst/stan/functions/obs_lpmf.stan
seabbs: We had a chat about this and the general conclusion is that I handled the normalisation of the delay distributions incorrectly but in a way that biased them vs recovering from simulation rather than impacting the nowcasts. Here they would also bias the nowcasts so time to change. 

The preferred solution was to lump all the truncated probability onto the maximum delay which mimics the handling of observations in the R code. We also thought it would make sense to do this as standard in the R code with no option to do anything else for now. 

 @adrian-lison if I am not mistaken you said you would take point on this?

inst/stan/functions/obs_lpmf.stan
seabbs: After having a chat we agreed I am going to try and implement dual likelihood functions with the missingness model in parallel across groups.

inst/stan/epinowcast.stan
seabbs: nope looks good.

inst/stan/epinowcast.stan
```diff
@@ -163,48 +175,92 @@ model {
...
+      for (i in (dmax+1):t) {
+        log_lik_miss[i - dmax] = 0;
+        for (k in 1:g) {
+          // log-likelihood for observations with missing reference date
+          log_lik_miss[i - dmax] += neg_binomial_2_lpmf(
+            obs_miss[k, i] | exp_obs_miss_rep[k][i], phi);
+        }
+      }
+    }
+    }
```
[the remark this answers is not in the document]
seabbs: ah yes good point.

inst/stan/epinowcast.stan
seabbs: Yeah I think changing this so they are all the same makes sense?

R/preprocess.R
seabbs: Opened an issue for this (#99)

R/preprocess.R
seabbs: We had a chat and agreed this is good to keep for now as likely to impact many users. If/when extending to arbitrary timespans this decision will need to be rethought.

inst/stan/epinowcast.stan
seabbs: Hmm so I think that we don't want to assume this for the general model but it seems like we need to for the missingness model? Given that I would guess we need to specify a check when a missingness model is asked for?

seabbs:
Note that after discussion we agreed I am going to split this into two PRs one for the data preprocessing and one for the model with the aim that the preprocessing PR will be merged into `0.1.0` and the model changes will be added to `0.2.0`.

seabbs:
This PR has been superseded by #106 and #107 as discussed.

inst/stan/functions/obs_lpmf.stan
adrian-lison: @seabbs Exactly, got this on my list

inst/stan/functions/obs_lpmf.stan
adrian-lison: @seabbs Yes, I stated this limitation with `dmax` also further below... The question is whether fitting the model on a time window shorter than the maximum delay is a sensible idea at all / whether it would work. With a parametric delay model - maybe... If you think it would be wortwhile to support this (e.g. beginning of new outbreak), then we can of course add a case distinction/min

PR #65, opened 2022-06-25
Switch to using neg_binomial_2_log

seabbs:
This PR attempts to stabilise and optimise the calculation of expected observations by increasing the proportion of the calculation performed on the log scale. This results in reduced computation time with the majority of this coming from switching to using the `neg_binomial_2_log` family of functions (over their natural scale counterparts).

I have tested this on the readme and found a speed-up of approximately ~ 15%. I haven't observed any negative side effects apart from a slight reduction in clarity. There is potentially more to be done here and I wonder if the repeated logit (and inverse logit) transforms could in some way be optimised.


inst/stan/functions/hazard.stan
```diff
@@ -20,13 +20,12 @@ vector cumulative_converse_hazard(vector h) {
...
-  ch = exp(ch);
   return(ch);
 }
 
-vector hazard_to_prob(vector h) {
+vector hazard_to_log_prob(vector h) {
   int l = num_elements(h);
   vector[l] p;
-  p[1:l] = h[1:l] .* cumulative_converse_hazard(h[1:l]);
+  p[1:l] = log(h[1:l]) + cumulative_converse_hazard(h[1:l]);
```
adrian-lison, line 29: I was only wondering whether we should also rename the `cumulative_converse_hazard` to `cumulative_converse_log_hazard` to indicate that it converts the hazard from unit to log scale.

adrian-lison (commented):
This is awesome! All looking great, one minor comment.
I might go and steal this for my dev model, I could imagine this being even more impactful if you have a renewal process (as the sampler explores large R values) ;)

inst/stan/functions/hazard.stan
seabbs: yeah, that is a good point.

seabbs: I'll make another PR.

seabbs: PR: #69

PR #70, opened 2022-06-28
Feature: Prototype logo

seabbs:
This PR contains a prototype logo as well as the code required to generate it. Currently, it is very minimal but will do for the moment.


seabbs (commented):
LGTM (self-review)

PR #80, opened 2022-06-29
Feature: Add metadelay

seabbs:
This PR adds a new helper function `enw_delay_metadata()`. This produces metadata about the delay distribution vector that may be helpful in future modelling. This prepares the way for #4 where this data frame with be combined with the reference metadata in order to build non-parametric hazard reference and delay based models. 

In addition to adding this function it has also been added to the output of `enw_preprocess_data()` in order to make the metadata readily available to end-users.


seabbs (commented):
LGTM (self-review).

PR #82, opened 2022-06-29
clarify use of enw_retrospective_data

### LINKED ISSUE #79: `enw_retrospective_data` potentially not working as expected
Second guessing as there is no function documentation but I'd expect

```r
retro_nat_germany <- enw_retrospective_data(
  nat_germany_hosp,
  rep_days = 30, ref_days = 40
)
```

to return a data set with 30 reporting days and 40 occurrence dates. Instead I get

```r
> length(unique(retro_nat_germany$report_date))
[1] 41
> length(unique(retro_nat_germany$reference_date))
[1] 41
```

It seems to me these two lines should both have a `>` sign, but perhaps I might be misunderstanding what is being done here
https://github.com/epiforecasts/epinowcast/blob/a5616b15857d790eced0f1289b498d42fa0ca6d7/R/preprocess.R#L171
https://github.com/epiforecasts/epinowcast/blob/a5616b15857d790eced0f1289b498d42fa0ca6d7/R/preprocess.R#L176

sbfnk:
Aims to fix #79 with documentation and clearer argument names.


R/preprocess.R
```diff
@@ -163,33 +165,44 @@ enw_add_max_reported <- function(obs) {
...
+      stop(
+        "`include_ref_days` and `earliest_rep_date` can't both be specified."
+      )
     }
-    earliest_ref_date <- max(retro_data$reference_date) - include_ref_days
+    earliest_ref_date <-
+      max(retro_data$reference_date, na.rm = TRUE) - include_ref_days
   }
-  retro_data <- retro_data[reference_date > earliest_ref_date]
+  retro_data <- retro_data[reference_date >= earliest_ref_date]
```
sbfnk, line 205: I thought we'd agreed on a strict greater than as per https://github.com/epiforecasts/epinowcast/issues/79#issuecomment-1169985877?

seabbs: We did indeed but that was not my read of the documentation as written now. Just bringing the two together before making a decision.

seabbs:
So I looked at this a bit more and thought that maybe it would be better if `enw_retrospective_data` was split into two functions and standardised with `enw_latest_data`

Doing this gives the following workflow which personally I think is a little clearer and easier to use (much easier if allowing yourself to use some kind of pipe). 

```r
# Load epinowcast and data.table
library(epinowcast)
library(data.table)

# Load and filter germany hospitalisations
nat_germany_hosp <- germany_covid19_hosp[location == "DE"][age_group %in% "00+"]
nat_germany_hosp <- nat_germany_hosp[report_date <= as.Date("2021-10-01")]

# Make a retrospective dataset
retro_nat_germany <-  nat_germany_hosp |>
  enw_filter_report_dates( remove_days = 40) |> 
  enw_filter_reference_dates(include_days = 40)

# Get latest observations for the same time period
latest_obs <- nat_germany_hosp |>
  enw_latest_data() |>
  enw_filter_report_dates( remove_days = 40) |> 
  enw_filter_reference_dates(include_days = 40)
```

The potential downside here is the duplication for retrospective and latest data but I think that might be hard to avoid.

PR #84, opened 2022-06-30
Feature: exponential delays

### LINKED ISSUE #83: Support exponential delay distributions
At the moment only lognormal and gamma delay distributions are supported. Given many example datasets exhibit near exponential reporting distributions and that this distribution depends on only a single parameter (and hence is easier to fit) it makes sense to support it. Note that this should in theory be similar to a model with a constant hazard (whilst being less efficient). This means that when this model is implemented it is unlikely that users will want to use the parametric approach. 

This requires:

- Updating the internal stan code to allow for `logsd` to be an optional parameter.
- Add support for estimating the exponential pdf.
- Add R code supporting passing initial conditions
- Add R code adding the exponential distribution to the user interface.

seabbs:
This PR closes #83 by adding support for exponential delays. It has no impact on default functionality. Checking the exponential delays posteriors they look reasonable. Checking posterior predictions indicates that performance is slightly worse than with the log-normal (at least on this very limited test set). The major issue seems to be capturing the first day of reporting which I find a little surprising.

README-week_nowcast-1

README-nowcast-1
README-pp-1


sbfnk (commented):
Looking good to me. Would it be worth reserving `dist == 0` for a model option without parametric delay distribution at all, i.e. the nonparametric components only?

Minor point: should update the comment in l.33 of `inst/stan/epinowcast.stan`.

inst/stan/epinowcast.stan
```diff
@@ -89,26 +90,30 @@ transformed parameters{
...
+        ref_lh[, i] = logit(ref_lh[, i]);
+      }
+    }else{
+      ref_lh = pmfs;
+    }
     }
-  }else{
-    ref_lh = pmfs;
-  }
   }
```
adrian-lison, line 117: I'm a bit confused: if dist==0 (exponential pmf), the reporting probs and ref_lh are never computed, are they? Also, why would we not want to support a reference day model for exponential delays? The hazard is only constant over the delays for a given reference date, but it should be able to change over the reference dates. Based on Sebastian's proposal I would have expected to have dist==0 and dist==exponential to be two different options.

seabbs:
I've updated this so that `dist = 0` corresponds with no parametric hazard model. At the moment that gives something fairly silly as the non-parametric reference day model is not in place but will be useful for the future. 

As part of these changes I've done a little bit of internal refactoring to make things easier (@adrian-lison 
I will add this in to #64 before we catch up on it).

seabbs:
Merging but we should circle back with a check before`develop` -> `main`

PR #85, opened 2022-07-01
compatibility with rstan for exposing functions

sbfnk:
Finalises the work in https://github.com/epiforecasts/epinowcast/pull/75#issuecomment-1171468409 to allow exposing all functions by:
- replacing `cdf` syntax from bar to comma
- replacing unnormalised `upmf`s with normalised ones
- moving variable declarations to the beginning of scopes
- avoiding overloading the custom zero truncated normal function by making `eobs_lsd` a vector

Also streamlines regex for arrays, and adds a newline to remove a warnings, and changes a filename to be consistent with others (underscore rather than dash).

With this the following code exposes all functions without error/warning.
```r
> function_files <- list.files(here::here("inst", "stan", "functions"))
> function_files <- c("hazard.stan", setdiff(function_files, "hazard.stan")) ## nead hazrds first as dependency
> expose_stan_fns(function_files, "inst/stan/functions")
```


seabbs (approved):
LGTM to me. I've added some docs and tests + extended the regex to a few more edge cases.

R/utils.R
```diff
@@ -4,36 +4,77 @@
...
+  # replace array syntax
+  #   case 1: array[] real x -> real[] x
   functions <- gsub(
-    "array\\[,\\] ([a-z]+) ([a-z_]+)", "\\1[,] \\2", functions
+    "array\\[(,?)\\] (.*) ([a-z_]+)", "\\2[\\1] \\3", functions
   )
+  #   case 2: array[n] real x -> real x[n]
   functions <- gsub(
-    "array\\[([a-z]+)\\] ([a-z]+) ([a-z_]+)", "\\2 \\3[\\1]", functions
+    "array\\[(.*)\\] (.*) ([a-z_]+)", "\\2 \\3[\\1]", functions
```
sbfnk, line 46: I think the first one is slightly risky as it will go greedily across lines. I'd suggest to replace `array\\[(.*)\\]` with `array\\[([^]]*)\\]` to make sure it stops at the first closing bracket.

sbfnk: If aiming to be as general as possible might also want to ensure it works with any amount of any whitespace by changing to something like
`"array\\[([^]]*)\\]\\s+(.*)\\s+([a-z_]+)", "\\2 \\3[\\1]"`

seabbs: hahaha I hate regex.

seabbs: https://github.com/epiforecasts/epinowcast/pull/93

PR #93, opened 2022-07-04
Bug safer regex

seabbs:
Safer regex as suggested by @sbfnk here: https://github.com/epiforecasts/epinowcast/pull/85#discussion_r913227210

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -43,7 +43,12 @@ convert_cmdstan_to_rstan <- function(functions) {
   )
   #   case 2: array[n] real x -> real x[n]
   functions <- gsub(
-    "array\\[([^]]*)\\]\\s+(.*)\\s+([a-z_]+)", "\\2 \\3[\\1]", functions
+    "array\\[([^]]*)\\]\\s+([a-z_]+)\\s+([a-z_]+)", "\\2 \\3[\\1]", functions
+  )
+  #   case 3: array[nl, np] matrix[n, l] x -> matrix[n, l] x[nl, np]
+  functions <- gsub(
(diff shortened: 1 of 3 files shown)


seabbs:
Hmm a problem for tomorrow. Looks like it isn't getting the last bracket. 

```r
`actual`:   "vector[n] x[nl, pq;" 
`expected`: "vector[n] x[nl, pq];"
```

R/utils.R
```diff
@@ -43,7 +43,7 @@ convert_cmdstan_to_rstan <- function(functions) {
   )
   #   case 2: array[n] real x -> real x[n]
   functions <- gsub(
-    "array\\[(.*)\\] (.*) ([a-z_]+)", "\\2 \\3[\\1]", functions
+    "array\\[([^]]*)\\]\\s+(.*)\\s+([a-z_]+)", "\\2 \\3[\\1", functions
```
sbfnk, line 46: missing a closing `]` before the last quotation mark

Also I think you need to replace `(.*)` with `([a-z_]*)` otherswise you get the same greedy matching issues as before.

seabbs: That won't account properly for vector variables will it?

sbfnk: I must be missing something. Can you give an example where it wouldn't?

seabbs: I've added another line of regex to crudely deal with this. I think perhaps should fix it as bugs occur now? Also added a few harder tests for things not currently in functions but that could be. See https://github.com/epiforecasts/epinowcast/pull/93/commits/5953dd18d7389499a84a218defe483212d4e04ba

PR #102, opened 2022-07-06
Replace group with .group

### LINKED ISSUE #99: Rename group variable to be more unique
The same goes for "old_group" in
https://github.com/epiforecasts/epinowcast/blob/a48ae0a9ee0652d8b1ffbf6220484ec9c1de2114/R/preprocess.R#L352
I'd suggest `.group` and `.old_group` as column names to avoid conflict.

_Originally posted by @sbfnk in https://github.com/epiforecasts/epinowcast/pull/64#discussion_r909642037_

adrian-lison:
This PR should fix #99 on the develop branch. All group, old_group and new_group variables have been prefixed with a dot.

### DIFF
--- a/R/model-design-tools.R
+++ b/R/model-design-tools.R
@@ -187,7 +187,7 @@ enw_add_cumulative_membership <- function(metaobs, feature) {
     cfeatures <- grep(cfeature, colnames(metaobs), value = TRUE)
     metaobs[,
       (cfeatures) := purrr::map(.SD, ~ ifelse(cumsum(.) > 0, 1, 0)),
-      .SDcols = cfeatures, by = "group"
+      .SDcols = cfeatures, by = ".group"
     ]
   }
   return(metaobs[])
--- a/R/model.R
+++ b/R/model.R
@@ -55,14 +55,14 @@ enw_obs_as_data_list <- function(pobs) {
   # format latest matrix
   latest_matrix <- pobs$latest[[1]]
   latest_matrix <- data.table::dcast(
-    latest_matrix, reference_date ~ group,
+    latest_matrix, reference_date ~ .group,
     value.var = "confirm"
   )
   latest_matrix <- as.matrix(latest_matrix[, -1])
 
   # get new confirm for processing
   new_confirm <- data.table::copy(pobs$new_confirm[[1]])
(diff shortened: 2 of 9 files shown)


seabbs (approved):
Nice stuff, it looks like you've got everything. The only thing that appears to be missing is that the script in `inst` that generates the examples needs updating (this is causing the scoring tests to fail at the moment (probably have some more tests using these examples as a way of checking we are reproducing past results as here obviously been helpful). 

I quickly added a check function to look for these variables, restyled after merging `develop`, and added some tests for the new check function. 

Once examples are updated and there is a news item for this PR very happy for this to be merged in.

NEWS.md
```diff
@@ -4,20 +4,21 @@ This is a major release and contains multiple breaking changes. If needing the o
...
 * Two new helper functions `enw_filter_reference_dates()` and `enw_filter_report_dates()` have been added. These replace `enw_retrospective_data()` but allow users to similarly construct retrospective data. Splitting these functions out into components also allows for additional use cases that were not previously possible. See [#82](https://github.com/epiforecasts/epinowcast/pull/82) by [@sbfnk](https://github.com/sbfnk) and [@seabbs](https://github.com/seabbs).
+* The internal grouping variables have been refactored to reduce the chance of clashes with columns in the data frames supplied by the user. There will also be an error thrown in case of a variable clash, making preprocessing more save. See [#102](https://github.com/epiforecasts/epinowcast/pull/102) by [@adrian-lison](https://github.com/adrian-lison) and [@seabbs](https://github.com/seabbs), which solves [#99](https://github.com/epiforecasts/epinowcast/issues/99).
```
seabbs, line 13: save -> safe

R/check.R
```diff
@@ -50,26 +50,26 @@ check_dates <- function(obs) {
...
   if (!is.null(obs$.group)) {
     stop(
-      ".group is a reserved variables and must not be present in the input
+      ".group is a reserved variable and must not be present in the input
        data"
     )
   } else if (!is.null(obs$.new_group)) {
     stop(
-      ".new_group is a reserved variables and must not be present in the input
+      ".new_group is a reserved variable and must not be present in the input
```
seabbs, line 67: 🧙🏻

adrian-lison:
To me as well. Here we go!

PR #103, opened 2022-07-06
Feature cumulative sum expectation

### LINKED ISSUE #98: Update the expectation random walk to use the cumulative sum formulation
It is more efficient to encode random walks using a cumulative sum formulation. There should be no downsides to this change.

This section of the code

https://github.com/epiforecasts/epinowcast/blob/38ab91013fe105935c11bf2f9c199a2fbf0c2110/inst/stan/epinowcast.stan#L131

needs to be updated to use the same approach as seen here. 

https://github.com/epiforecasts/epinowcast/blob/a48ae0a9ee0652d8b1ffbf6220484ec9c1de2114/inst/stan/epinowcast.stan#L141

seabbs:
This PR closes #98 by adding a reformulation of the expectation random walk as a cumulative sum. It also adds a basic compilation test for the model. This has the downside of adding time to tests but the upside of more thoroughly testing if the model compiles.


seabbs:
I am seeing a compilation error both locally and on the CI. This happens with all model versions though so I think is a breaking change from `cmdstan 2.30.0` which I haven't been able to get to the bottom of yet.

seabbs:
Standard compilation works as expected (surprisingly without passing `include_paths`): 

```r
cmdstan_model("inst/stan/epinowcast.stan")
```

Compiling with `enw_model()` fails with model being unable to find functions. 

```r
enw_model()
```

seabbs:
This also fails, 

```r
enw_model("inst/stan/epinowcast.stan", include = "inst/stan")
```

seabbs:
Looking at the issues for `cmdstanr` and `cmdstan` I see nothing related. I also see nothing related in the  `cmdstan 2.30.0` release notes.

seabbs:
Rolling back to `cmdstan 2.29.0` I see successful compilation using,

```r
cmdstanr::install_cmdstan(version = "2.29.0")
enw_model()
```

seabbs:
Shall I merge this and try and resolve the `cmdstan` issue for `develop`. Can you replicate it?

PR #106, opened 2022-07-07
Feature missing reference dates preprocessing

seabbs:
This PR is a partial replacement for #64  and includes work on the preprocessing components of #43 (i.e missing data).

Issue #43, opened 2022-04-07
Support for missing reference dates MVP


adrian-lison:
### Design motivation
Some datasets may include counts of cases with missing reference dates. For these cases, the date of reporting is known but the date of occurrence is for some reason unknown. It would be a valuable feature to support such data.

One way to account for such "missing data" in nowcasting is imputation during preprocessing. This is typically done by estimating the delay distribution from historical data and imputing reference dates by subtracting random draws of the delay distribution from the reporting date. Regression models can be used to estimate the delay distribution, accounting for changes over time, on weekdays, group effects etc. This however means that two potentially sophisticated models are involved (the imputation model and nowcasting model), and to quantify the overall uncertainty, a multiple imputation scheme is required which can be computationally demanding if the nowcasting model must be rerun. More importantly however, subtracting delays from reporting days does not necessarily reconstruct the underlying time series by reference date. This is the same issue as with back-projection approaches used in R estimation (see https://doi.org/10.1371/journal.pcbi.1008409).

It thus seems more favorable to support missing reference dates directly in the nowcasting model by implementing a generative model of missingness and adding a likelihood term for the cases with missing reference date. In the following, an MVP version for `epinowcast` of such a model is outlined.

### Changes to the interface (preprocessing)
1. Implement an artificial/simulated dataset with cases with missing reference dates (`NA`) for testing and showcasing purposes. Ideally, also add a real-world dataset with cases with missing reference dates.
2. Adjust `enw_preprocess_data` to accept entries where the `reference date` is `NA` and gather them into a vector `obs_miss` that is indexed by reporting day. Add `obs_miss` to the returned `enw_preprocess_data` object.
3. Adjust `enw_obs_as_data_list` such that the `obs_miss` in `pobs` are also returned in the data list for stan.

### Changes to the model
1. Add `int obs_miss[rd]` to the data block. This array will contain the number of cases with missing reference date reported on a certain reporting date.
2. Add `vector<lower=0,upper=1>[t] alpha[g]` to the transformed parameters block. This will represent the share of cases with (eventually) recorded reference date information that occurred on a given reference date for a given group. In turn, `1-alpha` will represent the share of cases with (eventually) missing reference date information.
[...]

PR #107, opened 2022-07-07
Add stan code changes for missing data model

seabbs:
This PR is a partial replacement for #64 and builds on top of #106 with the required stan model changes. It cannot be merged until #106 has been merged. It implements the stan code changes suggested in #43.


seabbs:
@adrian-lison just an FYI but the redesign work is getting a bit out of hand and leading to a fairly broad change in the names of variables (to make it easier to automate in the R code but then requires stan changes to keep consistency). That means this PR may be easier if we rebase once that is done and go directly for a formula interface implementation. 

That being said if you have some time and want to work on this the thing that will not need updating (and is the trickest bit) is adding the two likelihood fns (i.e the one currently in `develop` and the missing data one we need here that is parallel over groups).

adrian-lison:
As discussed, we need to make the model save for `dmax>=t`. Since literally _no_ observations with missing reference date can be used in the likelihood in such a case, it probably makes sense to check for this in the beginning (transformed data) and either 1) reject (aka please use the non-missingness model version or a longer time window) or 2) switch to the non-missingness model automatically, ignoring all missingness data and print a warning.
The good news is that in both cases we only need a simple check in the beginning and no complicated `min` functions in the missingness likelihood as it doesn't make sense anyway, so this is an easy task. Should be done once we have the two likelihoods and the optional switch implemented:

- [ ] Handle `dmax>=t` case.

adrian-lison:
Further ToDos
- [ ] Change missingness model parts to log scale
- [ ] Implement two different likelihood functions
- [ ] Implement matrix-based version for obs in missingness likelihood
- [ ] Implement group-wise within-chain parallelisation for missingness likelihood

seabbs:
Shall I do another conflicts PR for this and add in the new variables from `enw_missing()`?

seabbs:
Just noting I have started to fix the merge issues introduced in `0.1.0`. I will have something tomorrow I think.

seabbs:
had a go at this but may have done more harm than good. See #133 where thinking that doing some more work on modularising the observation model in `develop` may help in making this addition (and in making future additions easier).

seabbs:
I might role back the work I did on this yesterday as I was losing my mind it appears!

adrian-lison:
I am sorry this became such a mess. Might have not been the best timing with this PR vs. the other breaking changes.

adrian-lison:
Regarding the matrix-based version for obs in the missingness likelihood, I suggest to do it roughly like I did it here:
https://github.com/adrian-lison/nowcast-transmission/blob/main/code/models/model/likelihood_reported.stan

That is, have a matrix
`matrix[D, T] expected_obs_miss = rep_matrix(0, D, T);`

then, for every time `t` (assuming we are in one group at the moment)
`expected_obs_miss[1:min(T-t,D), t] = expected_obs_all + alpha1m_log[t];`

And then, after the loop, something like this:
```
  // reshape reporting triangle by date of report, leaving out left-truncated dates
  matrix[D, T-D] expected_obs_miss_rep
    = reporting_triangle_by_report(expected_obs_miss, D);
  // compute log expected cases with missing occurrence for each date of report
  vector[T-D] expected_obs_miss_rep_vec;
  for (i in 1:(T-D)) {
    expected_obs_miss_rep_vec[i] = log_sum_exp(col(expected_obs_miss_rep,i));
  }
```
The function `reporting_triangle_by_report` I have implemented here: https://github.com/adrian-lison/nowcast-transmission/blob/main/code/models/functions/reporting_triangles.stan (might need some tweaking because of the different max_delay definitions).

Then we can use `expected_obs_miss_rep_vec` in a vectorized negative binomial likelihood.

seabbs:
Closing in favour of #138 and #147

PR #111, opened 2022-07-11
Change prior transformation from inv_sqrt to inv_square

adrian-lison:
This PR changes the implementation of the prior on `phi` to what is probably intended: if we put a half-normal prior on `1/sqrt(phi)`, we have to transform the parameter using `inv_square`, not `inv_sqrt`, to obtain `phi`. I think this means that the current implementation had a larger prior on `phi` than intended, meaning less overdispersion than intended.

### DIFF
--- a/inst/stan/epinowcast.stan
+++ b/inst/stan/epinowcast.stan
@@ -128,7 +128,7 @@ transformed parameters{
   }
   }
   // transform phi to overdispersion scale
-  phi = inv_sqrt(sqrt_phi);
+  phi = inv_square(sqrt_phi);
   // debug issues in truncated data if/when they appear
   if (debug) {
 #include /chunks/debug.stan


adrian-lison:
Note that I get some divergent transitions when running the current model version from develop, but this does not seem to be related to the overdispersion prior (getting them with both versions).

seabbs:
Hmm, thanks for this Adrian good spot. I am not sure I totally understand? I thought `inv_sqrt` was doing $\frac{1}{\sqrt{phi}}$ which is, I think, what we had before and is recommended for overdispersion? 

Not a good sign on the divergent transitions and odd I haven't seen them. I wonder if this is due to the change in long delays in the R code vs in the stan code or something else? 

## Documentation

Screenshot 2022-07-11 at 09 42 15

https://mc-stan.org/docs/functions-reference/power-and-logarithm-functions.html

seabbs:
Based on BDA3 and this wiki: https://github.com/stan-dev/stan/wiki/Prior-Choice-Recommendations#story-when-the-generic-prior-fails-the-case-of-the-negative-binomial

adrian-lison:
Yes, exactly, I also think that `inv_sqrt` is doing `1/sqrt(x)`. But in the current model, we say `phi = 1 / sqrt(sqrt_phi)`, which is not how I understand the recommendation from the wiki. I read it as: we put a prior on `1/sqrt(phi)`. We call this parameter `sqrt_phi` in our implementation, hence `sqrt_phi = 1/sqrt(phi)`. Therefore, the transformation to obtain `phi` would have to be `phi = 1/(sqrt_phi^2) = inv_square(sqrt_phi)` I thought...

seabbs:
So I think you are correct (like 90%) from looking again and this does seem to follow from https://statmodeling.stat.columbia.edu/2018/04/03/justify-my-love/

I need to pull this in locally but before merging we also need to change the model description (perhaps we should have a reference for this prior choice to avoid digging again like this), check initialisation, and likely remove the upper bound on phi as just waiting compute I think (very small amounts but still).

adrian-lison:
I removed the upper bound from `sqrt_phi` and changed the initialisation for `phi`. By the way, do we even have to init `phi` when it is a transformed parameter?

adrian-lison:
> So I think you are correct (like 90%) from looking again and this does seem to follow from https://statmodeling.stat.columbia.edu/2018/04/03/justify-my-love/

Yeah, interestingly it is really hard to find examples on the web where people actually followed the recommendation exactly. Rstanarm seems to be doing it differently in general, epidemia too, [these](https://mc-stan.org/users/documentation/case-studies/boarding_school_case_study.html#complete_stan_program) people here only put a prior on `1/phi`, ... but I know a colleague has independently implemented it like me in his work [here](https://github.com/nbanho/npi_effectiveness_first_wave/blob/11c08e03ef3f2e4d124292ae4f18a973a53b5fd8/models/sim_g.stan#L58).

And I think it makes sense: if we "inverse" using `1/x`, we should also inverse the `sqrt` by squaring it.

seabbs (changes requested):
Looks good and I agree. Tested locally and works as expected. We just need: 

- [ ] News update
- [ ] Update to model documentation for this change (ideally with a reference)

seabbs:
>  do you mean with the model documentation

Sorry I meant the model vignette.

adrian-lison:
Divergent transitions were a false alert, I messed up when loading the current package version. Sorry for that one.

adrian-lison:
@seabbs I updated the vignettes, do you want to have a look? :)

And one more thing, how should we add a news item, now that the branch is based on main? Should we append this to 0.07 or add a 0.07.1?

adrian-lison:
Sorry for the confusion, I intended it for develop, but maybe I based on the wrong branch in the first place...

I reran the Germany vignette because I thought that the model outputs may have slightly changed with a different prior... But I can revert it if you think that makes sense.

seabbs:
Can we retarget this at develop without causing massive pain?

adrian-lison:
Yeah I think it should work. But changing the base branch seems to be something only you are allowed to do. I can resolve any conflicts after that

seabbs:
Done. I'll look into why you can't do that and try and fix.

seabbs (approved):
Looks good to me. Nice job sorting out the merge issues!

PR #112, opened 2022-07-11
Feature: Redesign interface

### LINKED ISSUE #57: Redesign interface
This is a proposal for a new interface. The basic idea is to build the model up as a series of chunks related to the sub-modules. There are open questions about how much of the data munging process (i.e which formulas use which data) the user should be exposed to. 

A possible candidate interface:

```
epinowcast(reporting = enw_reporting(...), expectation= enw_expectation(...), missing = enw_missing(), observation = enw_obs())

enw_reporting( 
  parametric = ~ 1,
  distribution = "log-normal", # | "none" = no parametric baseline hazard
  baseline = ~ 0 # (none) | ~ d (each day i.e discrete)
  report = ~ rw(week, by = age_group),
  reference = ~ day_of_week,
  structure = ~ saturdays # (known days reporting cannot happen and hazard to -Inf),
  data = obs
)

enw_expectation(
  ~ rw(date, by = age_group), 
  latent = FALSE # here could have for example "poisson"
  delay_to_report = 1  # here could have the delay from inf to report as a pmf. Later extend to uncertain dists
  generative_process = "log" # i.e as currently could also add options such as "renewal" etc
  data = obs
)

enw_missing(
   ~ 1, 
   data = obs
)

enw_obs(
  baseline = ~ 1 # we could offfer the ability to model the variance parameter if present by baseline (i.e delay,) report and reference date
  report = ~ 1
  reference = ~ 1
  family = "negbin" #other options could be Poisson etc or even potentially non-count options. 
)
```

seabbs:
This PR aims to resolve #57. See the issue for interface design details and comments below for other changes.


seabbs:
Currently, something I am a bit stuck on is the interface for priors. At the moment this is centralised and it would be nice if it wasn't. I think for now I am going to propose sticking with the current system as I don't really have a good idea but this does mean a future release will need to also have breaking changes. Once the dust has settled a bit if there is still no good solution will open an issue.

seabbs:
This now implements everything discussed in #57 but I have yet to add documentation for the new functions or testing so it isn't quite ready for review. 

That being said as it feature complete it would be good to get feedback at this stage. In particular on the setup of the modularity. This is all in `R/model-modules.R` and is linked together in `R/epinowcast.R`. At the moment everything is implemented with a simple list structure but potentially it would be useful to build out a s3 class etc but I think this can be done later/during another dev pass without breaking changes. 

Something that is currently less than ideal is the passing of priors, in particular you cannot set priors for fixed effects. This makes things a bit clunky. In other work (https://github.com/seabbs/epict/blob/main/R/model.R and https://github.com/seabbs/epict/blob/a462b2482d2ddd176085e12ed824b48c42eb1a26/R/extract.R#L59) I've done a bit more on this and also got linking with model output working. I am still not happy with that implementation so planning on punting this to a later PR (which may sadly add some breaking changes but hopefully to less used functionality).

seabbs:
In https://github.com/epiforecasts/epinowcast/pull/112/commits/50133cbb0e26a1fe5812a38336d14e893f7eed50 I fixed a small bug introduced in #83 which incorrectly if gated the reference/report models meaning that the simplest nowcasting model could not be run as expected.

seabbs:
This is now documented and passing CRAN check. The remaining work to do is: 

- [ ] Add news update
- [ ] Add tests for new model tools
- [ ] Add tests for each model module
- [ ] Add non-sampling tests for `epinowcast` (by using a dummy sampler function)

seabbs (commented):
LGTM apart from linting flagged test gaps (self-review)

PR #113, opened 2022-07-12
Update discretised_reporting_prob

### LINKED ISSUE #109: Update discrete_reporting_prob
After discussion (between @adrian-lison  and myself) we concluded that the discrete reporting probability should be normalised to 1 by putting 1 minus the cumulative probability density on the last permissible max delay rather than the current default of normalising.  This should be added as an option in the stan code but hard coded in the actual model as no other use case makes sense currently.

Before merging we need to fit against the example and README to check performance.

adrian-lison:
This PR closes #109 by adding three options for the discretised reporting probability: ignore delays beyond max, add to max, or normalize. The option "Add to max" is used as the only/default option in the epinowcast model for now.

No changes to preprocessing seem required as the add_to_max strategy was made default by an earlier PR.

Model tested locally and runs correctly.

One thing to potentially address here before merging (aside from adding to NEWS, ToDo @adrian-lison) is to ensure that the handling of max delay is communicated in an unambigious way everywhere. My understanding of the current implementation is: if we set a max delay of 5, then we model the following delays 0, 1, 2, 3, 4. Everything beyond (5 and later) is added to delay 4. I am not 100% sure yet if this is intuitive: on the one hand it is true that we model 5 delays, on the other hand users might expect that if the max delay is 5, we model delays up to and including five days after occurrence. Above all, our decision should be clearly communicated, so it could be worthwile to go through the function documentations etc. again. I also think that the preprocessing and visualisation of the estimated and empirical delays will help a lot later.


seabbs (approved):
This looks good. 

It occurs to me looking at this that perhaps we can do more on the log scale here making use of log_exp_sum and friends. It isn’t clear to me this would be more efficient/stable but I think it could be. 

Totally agree that the documentation is ambiguous and the current definition for the maximum delay  may not be the one users expect. Definitely think worth having an issue and trying to get a range of opinions.  Perhaps wait for that vs covering in this PR?

In #112 I am adding a helper function for the reference date model. Perhaps this would be a good place for a very clear discussion of parametric modelling assumptions. The other obvious targets are the model definition vignette and the preprocessing docs.

adrian-lison:
Great, added a bullet points to the News. Also added #114 to decide on the max delay issue later.

Regarding the use of `log_sum_exp`, I am not yet sure. I am currently experimenting with widespread use of it in my R nowcasting model (all the convolutions become `log_sum_exp`) and can share insights soon.

seabbs:
> can share insights soon

Keen to hear what you find. 

Happy to see this PR merged when you are ready.

Issue #105, opened 2022-07-07
Compilation issue with stan 2.30.0


seabbs:
I am seeing a compilation error both locally and on the CI when using `develop` and  in #112 . This happens with all model versions so I think is a breaking change from `cmdstan 2.30.0` which I haven't been able to get to the bottom of yet rather than a model-specific issue. 

Confirmation if others are seeing this would be useful. If this issue impacts `main` as well (which I think likely) then any fix needs to be targetted there as well. 

_Originally posted by @seabbs in https://github.com/epiforecasts/epinowcast/issues/103#issuecomment-1176733477_

seabbs:
This also fails, 

```r
enw_model("inst/stan/epinowcast.stan", include = "inst/stan")
```

seabbs:
Looking at the issues for `cmdstanr` and `cmdstan` I see nothing related. I also see nothing related in the  `cmdstan 2.30.0` release notes.

seabbs:
Rolling back to `cmdstan 2.29.0` I see successful compilation using,

```r
cmdstanr::install_cmdstan(version = "2.29.0")
enw_model()
```

seabbs:
To resolve this in the CI `cmdstan` has been pinned to `2.29.0` any updates looking to fix this should roll this back.

seabbs:
I still haven't gotten to the bottom of this.

sbfnk:
For me,
`enw_model(stanc_options = list())` 
works while
`enw_model()`
yields an error so it seems to be related to the optimisation (looks like a `cmdstan` error to me)

seabbs:
ah, that is good to know. For now, we can just drop passing optimisation as it only gave a very small speed up anyway.

seabbs:
Done in #117

PR #117, opened 2022-07-15
Fix compilation issue from cmdstan 2.30.0

seabbs:
This PR fixes the compilation issue flagged in #105 using the suggestion from @sbfnk to drop optimisation flags. Tested locally everything now works as expected. In addition to the above change stan code has been updated to remove warnings from cmdstan `2.30.0` and CI has been reset to use the latest `cmdstan` version.

### DIFF
--- a/R/model.R
+++ b/R/model.R
@@ -287,10 +287,6 @@ enw_inits <- function(data) {
 #' @param profile Logical, defaults to `FALSE`. Should the model be profiled?
 #' For more on profiling see the [`cmdstanr` documentation](https://mc-stan.org/cmdstanr/articles/profiling.html). # nolint
 #'
-#' @param stanc_options A list of options to pass to the `stanc_options` of
-#' [cmdstanr::cmdstan_model()]. By default "01" is passed which specifies simple
-#' optimisations should be done by the prior to compilation.
-#'
 #' @param ... Additional arguments passed to [cmdstanr::cmdstan_model()].
 #'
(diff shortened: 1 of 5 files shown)


R/model.R
```diff
@@ -287,10 +287,6 @@ enw_inits <- function(data) {
 #' @param profile Logical, defaults to `FALSE`. Should the model be profiled?
 #' For more on profiling see the [`cmdstanr` documentation](https://mc-stan.org/cmdstanr/articles/profiling.html). # nolint
 #'
-#' @param stanc_options A list of options to pass to the `stanc_options` of
```
sbfnk, line 290: might want to keep the description and just update the default bit? (and re-render the documentation)

tests/testthat/test-enw_model.R
```diff
@@ -1,14 +1,19 @@
 test_that("enw_model can compile the default model", {
   skip_on_cran()
-  enw_model()
+  expect_error(enw_model(), NA)
```
sbfnk, line 3: do we?

seabbs: This is the hacky way to test it doesn't error.

R/model.R
seabbs: Done in [bfdf487](https://github.com/epiforecasts/epinowcast/pull/117/commits/bfdf487d585d8887865297961502258b3f0cccc7)

PR #128, opened 2022-07-19
Add loglogistic parametric distribution

adrian-lison:
This PR adds the loglogistic distribution as a fourth option for the parametric delay distribution. It is similar to the lognormal (slightly heavier tails), and often used in survival modeling due to its closed-form solution for the hazard function. I have managed to work out the closed-form solution for the discrete hazard case, so this will be another parametric function (aside from the exponential) that we can efficiently implement on the hazard scale (i.e. using vectorized, simple computations only). Can work on this after this PR is merged.

### DIFF
--- a/R/model-modules.R
+++ b/R/model-modules.R
@@ -10,7 +10,7 @@
 #'
 #' @param distribution A character vector describing the parametric delay
 #' distribution to use. Current options are: "none", "lognormal", "gamma",
-#' and "exponential" with the default being "lognormal".
+#' "exponential", and "loglogistic", with the default being "lognormal".
 #'
 #' @param non_parametric A formula (as implemented in [enw_formula()])
 #' describing the non-parametric logit hazard model. This can use features
@@ -40,7 +40,7 @@ enw_reference <- function(parametric = ~1, distribution = "lognormal",
     stop("The non-parametric reference model has not yet been implemented")
   }
   distribution <- match.arg(
-    distribution, c("none", "exponential", "lognormal", "gamma")
+    distribution, c("none", "exponential", "lognormal", "gamma", "loglogistic")
   )
   if (distribution %in% "none") {
     warning(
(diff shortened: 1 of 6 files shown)


R/model-modules.R
```diff
@@ -10,7 +10,7 @@
 #'
 #' @param distribution A character vector describing the parametric delay
 #' distribution to use. Current options are: "none", "lognormal", "gamma",
-#' and "exponential" with the default being "lognormal".
+#' "exponential", and "loglogistic", with the default being "lognormal".
```
seabbs, line 13: should we link out to distribution definitions here do you think? I particularly wonder if the different definitions of summary parameters will confuse people. Perhaps this is something for another issue.

adrian-lison: Yes, I agree, will make a new issue for this. Definitely important since they also translate differently into the moments of the distributions.

inst/stan/functions/discretised_reporting_prob.stan
```diff
@@ -17,6 +17,12 @@ vector discretised_reporting_prob(real mu, real sigma, int n, int dist, int max_
     for (i in 1:n) {
       upper_cdf[i] = gamma_cdf(i | emu, sigma);
     }
+  } else if (dist == 4) {
+    real emu = exp(mu);
+    real esigma = exp(sigma);
```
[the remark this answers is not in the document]
adrian-lison: Good catch!

adrian-lison:
@seabbs Do you want me to add a test with this distribution? And just whether the model runs or also including the fit?
I am only worried that having too many model integration tests will slow down the testing a lot. Maybe it is only worth for features that we expect to interact strongly with other future additions? Or are there tests that we can set to only run if triggered manually?

seabbs:
No not for this addition but I think at some point we should consider adding more unit tests of the underlying stan functions.

seabbs (approved):
Looks ready to merge to me.

Issue #118, opened 2022-07-15
Links in NEWS.md


seabbs:
In order to track development, we have been using a news file where each change/feature is described using a bullet point. In these bullet points we generally link to the issue (using the #number system used by GitHub) and highlight who contributed to the PR (using the @author GitHub tag). Up until now we have manually created markdown links for these but GitHub already automatically does this. It turns out that `pkgdown` which is the other entry point to our documentation also does this automatically. 

This means that we could streamline our documentation by removing the markdown links and allowing GitHub/`pkgdown` to do this for us. It does mean that users reading our NEWS.md locally or from another hosting provider won't have links so potentially worth considering if we want to do this.

choi-hannah:
Hi I'm happy to help with this

seabbs:
Amazing, thanks! The first step is to have a look at the [contributing guide](https://github.com/epiforecasts/epinowcast/blob/main/CONTRIBUTING.md). Give me a ping if you come across any issues.

seabbs:
Closed in #132 by @choi-hannah 🥳

PR #132, opened 2022-07-22
Feature improve news

choi-hannah:
This PR closes issue #118 which was to remove explicit links to authors and issues. I have also added .Rhistory to the gitignore. Note: I've excluded one of them because it linked to a specific comment in the issue.

I have read the contributing guide and I have checked this contribution locally.

Note that I saw failures were running devtools::check but this was related to the stan model. Because my changes were unrelated to this, it's probably fine.


NEWS.md
```diff
@@ -1,7 +1,7 @@
 # epinowcast 0.2.0
 
 ## Model
-- Added support for parametric log-logistic delay distributions. See [#128](https://github.com/epiforecasts/epinowcast/pull/128) by [@adrian-lison](https://github.com/adrian-lison).
+- Added support for parametric log-logistic delay distributions. See #128 by @adrian-lison.
```
seabbs, line 4: Missing a news entry for this PR. Ideally, we need a comment saying what you have done and then linking to the PR and PR author (as done by Adrian here). 

As this is a package change can you add a ## Package subheading for your new entry as we did for `epinowcast 0.1.0`

seabbs (changes requested):
Thanks for this @choi-hannah this looks great - thanks for the contribution! 

Just missing a news entry for this PR - so once that is added and CI checks are passing request a rereview and we should be okay to merge. 

Note that this would be a contributor-level addition for us so if you would like to be added to the DESCRIPTION as a `ctb` please do that in this PR.

NEWS.md
choi-hannah: I have done this

seabbs (approved):
Thanks looks good now and thanks for fixing the DESCRIPTION indenting!

Just waiting for the CI checks to complete and then will merge 🥳

PR #134, opened 2022-07-22
Hazard specification of parametric baseline hazard

### LINKED ISSUE #56: Hazard specification of parametric baseline hazard
At the moment the parametric baseline hazard is calculated by using the CDF of the distributions to get the discrete probability of report. It may be possible to instead directly specify this as a hazard. This may be optimal when combined with other hazard models (like proportional changes in hazard over time etc).

adrian-lison:
This PR closes #56 by implementing direct computation of hazards for the parametric delay distributions. Only in the case of no reporting effects the probabilities are instead computed (same case distinction as before).

The model fits correctly, but speed is improved only marginally (on the order of 0.1% compared to the likelihood). This may be bigger for larger delays, and anyway I think this is the cooler way to specify the baseline 😆 

FYI: Initially, I considered adding special cases/shortcuts for the exponential and log-logistic distribution. However, I found that no relevant improvement of speed would be achieved at the cost of making the code less elegant. In particular, the exponential distribution only has a constant hazard _if we don't normalize it_, so it's not worth implementing a case distinction for this.


inst/stan/epinowcast.stan
```diff
@@ -86,14 +86,14 @@ parameters {
 transformed parameters{
   vector[refp_fnrow] refp_mean;
   vector[refp_fnrow] refp_sd;
-  matrix[dmax, refp_fnrow] pmfs; // sparse report distributions
```
adrian-lison, line 89: I threw out the pmfs and did not replace it with the hazards - could be changed if having the hazards as a transformed parameter is important.

inst/stan/functions/discretised_reporting_prob.stan
```diff
@@ -25,21 +23,53 @@ vector discretised_reporting_prob(real mu, real sigma, int n, int dist, int max_
...
+  vector[n] haz; 
+  vector[n] cdf;
+  vector[n] ccdf;
+  cdf = upper_cdf_discrete(mu, sigma, n, dist);
+  cdf = adjust_cdf_discrete(cdf, n, max_strat);
+  ccdf = 1 - cdf;
+  // compute discretised hazard
+  haz[1] = cdf[1];
+  haz[2:(n-1)] = 1 - ccdf[2:(n-1)]./ccdf[1:(n-2)];
+  haz[n] = 1;
```
adrian-lison, line 73: Just like in the `prob_to_hazard` function, we are forced to set `haz[n]=1` as otherwise numerical issues can arise (non-finite gradients).

Come to think about it, we are currently not able to actually implement `max_strat=0`, because we have to set the last hazard to 1 (making it `max_strat=1` effectively). So the question is whether we should remove it or just leave it there. It doesn't do much harm at least.

inst/stan/chunks/debug.stan
```diff
@@ -1,17 +1,17 @@
     for (i in 1:refp_fnrow) {
       int j = 0;
       for (k in 1:dmax) {
-        j += is_nan(abs(pmfs[k, i])) ? 1 : 0;
-        j += is_inf(abs(pmfs[k, i])) ? 1 : 0;
+        j += is_nan(abs(ref_lh[k, i])) ? 1 : 0;
+        j += is_inf(abs(ref_lh[k, i])) ? 1 : 0;
```
adrian-lison, line 5: Replaced `pfms` with `ref_lh`. Could be `hazards` if we want to keep them in `transformed parameters`.  Btw, why do we use `abs` here (and `cmdstan` is asking to change it to `fabs`)?

adrian-lison:
Still need to sort out the tests apparently...

inst/stan/chunks/debug.stan
seabbs: In 2.30.0 it warns about fabs. In 2.29.0 it warns about abs. I think the flip flopped on which to implement. I saw some issues to this effect but now can't find.

inst/stan/epinowcast.stan
seabbs: No don't think, so this is sensible.

inst/stan/epinowcast.stan
```diff
@@ -103,20 +103,20 @@ transformed parameters{
...
-         2);
-    }
-    if (ref_as_p == 0) {
+    // calculate reference date logit hazards (unless no reporting effects)
+    profile("transformed_delay_reference_date_hazards") {
+    if (ref_as_p) {
       for (i in 1:refp_fnrow) {
-        ref_lh[, i] = prob_to_hazard(pmfs[, i]);
-        ref_lh[, i] = logit(ref_lh[, i]);
+        ref_lh[, i] =
```
seabbs, line 110: Having this as an option adds a fair bit of model complexity. I like it for the speed up it gives but could decide it is too much technical debt to carry around.

inst/stan/functions/discretised_reporting_prob.stan
```diff
@@ -1,7 +1,5 @@
-// Calculate the daily probability of reporting using parametric
-// distributions up to the maximum observed delay
-vector discretised_reporting_prob(real mu, real sigma, int n, int dist, int max_delay_strat) {
-  vector[n] pmf; 
+// Compute the cdf of a parametric distribution at values 1:n
+vector upper_cdf_discrete(real mu, real sigma, int n, int dist) {
```
seabbs, line 2: I really love the change to make this more functional. 10/10

inst/stan/functions/discretised_reporting_prob.stan
```diff
@@ -25,21 +23,53 @@ vector discretised_reporting_prob(real mu, real sigma, int n, int dist, int max_
...
+  cdf = adjust_cdf_discrete(cdf, n, max_strat);
+  // compute discretised pmf
+  pmf[1] = cdf[1];
+  pmf[2:n] = cdf[2:n] - cdf[1:(n-1)];
   return(pmf);
 }
+
+// Calculate the daily hazard of reporting using parametric
+// distributions up to the maximum observed delay
+vector discretised_reporting_hazard(real mu, real sigma, int n, int dist, int max_strat) {
```
seabbs, line 63: This looks good - love it regardless of speed up. Same wonder as before about doing some more of this on the log scale but can save that for another time/discussion.

inst/stan/functions/discretised_reporting_prob.stan
seabbs: Yeah this is a good point. It also introduces different behaviour depending on if you use hazards or probability directly. As we don't surface the stan functionality I would suggest we leave in but with a comment noting this issue?

seabbs (approved):
Love this so much regardless of speed up! Everything looks good aside from failing tests.

seabbs:
I've merged in modular changes from #137 and updated the tests and examples. Seeing a little bit of numerical instability potentially?

inst/stan/functions/discretised_reporting_prob.stan
seabbs: See #140 for more wondering.

inst/stan/functions/discretised_reporting_prob.stan
seabbs: Hmm perhaps this is related to what I am seeing in #140 with infinite gradients. Not entirely clear if it can be as this is `Inf` on the logit scale.

seabbs:
Tested this out and no instability. Going to merge but lets keep discussing points above.

inst/stan/functions/discretised_reporting_prob.stan
adrian-lison: I think it could be the same issue.

Regarding commenting the limitation of `max_strat==0` do you want to add something to #140 now that this PR is merged? :)

inst/stan/chunks/debug.stan
adrian-lison: Oh I didn't know they were switching around. My question was more about why we need the absolute here at all, why can't we just plug in the potentially negative values into `is_inf`? I am probably lacking some knowledge on `is_inf` here?

inst/stan/epinowcast.stan
adrian-lison: I think you solved it elegantly in #140

adrian-lison:
Thanks for going the last mile @seabbs, added some comments, discussion continues on #140 I guess :)

inst/stan/chunks/debug.stan
seabbs: uummmmm I don't know what past Sam was thinking here. Maybe lack support for negatives? I have no idea. Will test.

inst/stan/functions/discretised_reporting_prob.stan
seabbs: Will do - sorry got excited!

PR #137, opened 2022-07-25
Feature modularise observations

### LINKED ISSUE #133: Modularise observation model
In trying to fix merge issues with #106 I found that it was quite difficult due to the amount of repeated code across functions. To make ongoing development easier I think we should look at making the observation model parts of the code more modular. 

This includes: 

- [x] Combining the logit hazards
- [x] Making a true obs function with vectorised and non-vectorised and rng options (precisely how to do this without code duplication is a little tricky)
- [x] Share more code across generated quantities and the main likelihood
- [x] Make a function for producing nowcasts from same data format as used in the main likelihood
- [x] Make a function for producing posterior predictions from the same format as used in the main likelihood
- [x] A vectorised version of `expected_obs()` that can be reused across functions/use cases.

seabbs:
This PR closes #133 by modularising some areas of the model code. This includes: 

- The observation model options are now contained in an overloaded function
- The logit hazards are now combined once in their own function rather than in `expected_obs()`
- Priors for regression effects are now set using a function to reduce repeated code blocks and make adding new modules easier. 

I've rerendered the README and found no functional change in performance or results as should be expected for a primarily style focussed refactor. We could potentially modularise further but may be easier to do so when/if needed as more functionality is added. 

I've tagged multiple reviewers but only one is needed.


adrian-lison (approved):
Really good work! This improves the readibility of the model code a lot. LGTM

seabbs:
I was planning to wait to merge this until #134 was merged in as assumed would cause some conflicts.

PR #140, opened 2022-07-26
Feature logit reference hazard

seabbs:
This PR builds on #134 by specifying the discretised reporting hazard/probability on the log scale directly in order to reduce run times/increase robustness. It also adds tests of the updated stan code, increases the modularity of the stan code, and fixes a small bug with `ref_as_p` which prevented the probability mode of the model being used (this needs to be added to `develop` regardless of what happens with this PR). 

Currently, these changes work as expected when tested function by function, and work as expected when the probability-only model is used (i.e no hazard effects) but fail (with a non-finite gradient at initialisation) when the logit hazard mode is used. 

I have yet to get to the bottom of what is driving this. It may be the presence of `positive_infinity()` or it may be another issue. Using print debugging everything appears to be as expected in the output during fitting.


inst/examples/germany_dow.R
```diff
@@ -46,7 +51,7 @@ nowcast <- epinowcast(pobs,
   fit = enw_fit_opts(
     save_warmup = FALSE, pp = TRUE,
     chains = 2, threads_per_chain = 1,
-    iter_warmup = 500, iter_sampling = 500
+    iter_warmup = 500, iter_sampling = 500, debug = TRUE
```
adrian-lison, line 54: Reminder to remove debug mode in the end?

inst/stan/chunks/debug.stan
```diff
@@ -2,7 +2,9 @@
       int j = 0;
       for (k in 1:dmax) {
         j += is_nan(abs(ref_lh[k, i])) ? 1 : 0;
-        j += is_inf(abs(ref_lh[k, i])) ? 1 : 0;
+        if (k < dmax) {
+          j += is_inf(abs(ref_lh[k, i])) ? 1 : 0;
+        }
```
adrian-lison, line 7: Haha makes sense

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -0,0 +1,89 @@
...
+
+// compute discretised logit hazard
+vector lcdf_to_logit_hazard(vector lcdf, int n) {
+  vector[n] lhaz;
+  vector[n] lccdf;
+  lccdf = log1m_exp(lcdf);
+  lhaz[1] = lcdf[1];
+  lhaz[2:(n-1)] = log1m_exp(lccdf[2:(n-1)] - lccdf[1:(n-2)]);
+  lhaz[1:(n-1)] = lhaz[1:(n-1)] - log1m_exp(lhaz[1:(n-1)]);
+  lhaz[n] = positive_infinity();
```
adrian-lison, line 69: Maybe you can try `lhaz[n] = logit(1);` just for a test?

adrian-lison (commented):
I could only do a code review (no time for running the code myself atm, sorry), but it all looks good to me implementation-wise, well done. Very imaginable that the infinite gradient issues are the same ones as I received. I also printed the intermediate outputs and could find no difference (both times the hazard at the final day was 1). I added a suggestion to try out in the comments below

inst/stan/functions/discretised_logit_hazard.stan
seabbs: yeah, I have tried a few variations of this and I get the issue in all of them (including converting everything back to the natural scale and doing it as you did originally). Really odd and not clear why. The only thing I have left to try I think is moving the `logit` out of the function and into the core stan code. Not at all clear why that would make a difference though.

seabbs: So I rolled back the hazard part to your version in https://github.com/epiforecasts/epinowcast/pull/140/commits/17378861a058f6c92babdbf6f1c42e7489f05b5a

This works as expected. In the next couple of commits will iterate on this to find what causes the infinite gradient.

seabbs: As soon as you switch the ccdf calculation to `exp(log1m_exp(lcdf))` then infinite gradients become an issue. Could be because `log1m_exp` is unstable when `lcdf` is near/equal to 1? (as results in `-Inf`)?

seabbs: https://github.com/epiforecasts/epinowcast/pull/140/commits/d992d5f19db4637c73ed88b603ebc3698398004b restricts the `ccdf` calculation to exclude the maximum (as known to be 1) and this now works as expected.

seabbs: The change to not calculate the maximum `lccdf` was suffiicent. https://github.com/epiforecasts/epinowcast/pull/140/commits/e6efb586329031fcd86c74f85b74cc77f3e0cfcb works as expected. This does make me concerned this may be unstable in settings where the maximum delay is much greater than the non-zero elements of the parametric delay.

seabbs: I constructed an extreme version of this scenario with no reporting beyond 10 days and a maximum delay of 20 days. It throws a few extra exceptions at initialisation but otherwise works well. 


```
# Load epinowcast and data.table
library(epinowcast)
library(data.table)

# Load and filter germany hospitalisations
nat_germany_hosp <- germany_covid19_hosp[location == "DE"][age_group %in% "00+"]
nat_germany_hosp <- enw_filter_report_dates(
  nat_germany_hosp,
  latest_date = "2021-10-01"
)

# Make sure observations are complete
nat_germany_hosp <- enw_complete_dates(
  nat_germany_hosp,
  by = c("location", "age_group")
)

# Exclude reported data beyond 10 days
nat_germany_hosp[report_date >= (reference_date + 10), confirm := NA]
nat_germany_hosp[,
 confirm := nafill(confirm, "locf"),
 by = c("reference_date")
]
# Make a retrospective dataset
retro_nat_germany <- enw_filter_report_dates(
  nat_germany_hosp,
  remove_days = 40
)
retro_nat_germany <- enw_filter_reference_dates(
  retro_nat_germany,
  include_days = 40
)

# Get latest observations for the same time period
latest_obs <- enw_latest_data(nat_germany_hosp)
latest_obs <- enw_filter_reference_dates(
  latest_obs,
  remove_days = 40, include_days = 20
)

# Preprocess observations
pobs <- enw_preprocess_data(retro_nat_germany, max_delay = 20)

# Reference date model
reference_module <- enw_reference(~1, data = pobs)

# Report date model
report_module <- enw_report(~ (1 | day_of_week), data = pobs)

# Compile nowcasting model
model <- enw_model(threads = TRUE)

# Fit nowcast model and produce a nowcast
# Note that we have reduced samples for this example to reduce runtimes
nowcast <- epinowcast(pobs,
  reference = reference_module,
  report = report_module,
  fit = enw_fit_opts(
    save_warmup = FALSE, pp = TRUE,
    chains = 2, threads_per_chain = 2,
    iter_warmup = 500, iter_sampling = 500
  ),
  model = model,
)

plot(nowcast, latest_obs)

plot(nowcast, type = "posterior") +
  facet_wrap(vars(reference_date), scale = "free")
```
[nowcast.pdf](https://github.com/epiforecasts/epinowcast/files/9198904/nowcast.pdf)
[plot.pdf](https://github.com/epiforecasts/epinowcast/files/9198905/plot.pdf)

seabbs:
CI issue may be related to this `rstan` issue: https://github.com/stan-dev/rstan/issues/954

seabbs:
This is ready to merge I think if we want to. I am not clear it adds actual stability/robustness as no real evidence of decreased runtimes or reduced errors etc. there is also the potential issue identified with very low probabilities leading to infinite gradients (though not observed with an extreme test: https://github.com/epiforecasts/epinowcast/pull/140#discussion_r931025243) 

That being said this is I suppose a more "correct" implementation and so on the strength of that I guess I would suggest we go with it and iterate on later? 

Note in https://github.com/epiforecasts/epinowcast/pull/140/commits/49f758b5b2ef8622207dcceaf8e9ab6c92076644 I added comments for `max_strat = 1` as discussed in #134.

adrian-lison (approved):
Nice fix (funny, we didn't even need the last ccdf), good to go from my side as well!

PR #142, opened 2022-07-26
Bug: ref_as_p not activating

### LINKED ISSUE #141: Probability only model not activating as expected.
The probability-only model (i.e only a parametric distribution is used and hence the hazard scale is not needed) is not used currently to a mistake specifying `ref_as_p` in the stan code. There is an additional issue in that the `enw_report()` module currently self-declares as on regardless of it is or not. 

Stan code issue: 

https://github.com/epiforecasts/epinowcast/blob/603782bece7cea70d68287d4481fc0dc08a006b8/inst/stan/epinowcast.stan#L68

`enw_report()` issue:

https://github.com/epiforecasts/epinowcast/blob/603782bece7cea70d68287d4481fc0dc08a006b8/R/model-modules.R#L182

seabbs:
This PR fixes #141 by correctly setting `ref_as_p` based on `model_rep` and `model_refp` as passed by `enw_report()` and `enw_reference()`. It also adds additional tests to cover this issue in the future. 

It replicates the fix added in #140 and replaces it.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -9,9 +9,12 @@
 - Implemented direct specification of parametric baseline hazards. See #134 by @adrian-lison.
 - Refactored the observation model, the combination of logit hazards, and the effects priors to be contained in generic functions to make extending package functionality easier. See #137 by @seabbs.
 
-
 ## Documentation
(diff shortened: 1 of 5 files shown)


seabbs (commented):
LGTM (self-review)

PR #147, opened 2022-07-31
Add missing reference model components to delay_group_lmpf and generated quantities

seabbs:
As pointed out in #138 I have now made things a bit complicated for us (@adrian-lison and myself) in terms of implementing the missing reference model. This PR aims to implement a simplistic version of the missing reference model to the `delay_group_lmpf ` and to the generated quantities with the idea we can make an improved version in the future/based on this implementation. 

So far I have implemented: 

- An if else option in `delay_group_lpmf` based on if the missingness model is required
- An allocation function to go from all (i.e up to `dmax`) observations to just observed observations
- An allocation function to apply missing reference effects to log observations
- A function to map (using a lookup) from observations/expectation by report and reference to observations/expectations by report. 

Before being marked as not a draft this PR needs: 

- [x] A lookup to be implemented in `enw_missing()`
- [x] Generated quantities to be updated with missing model
- [x] Filter for only observations with complete reports by reference date (efficiently)
- [x] The addtion of a basic regression test for the missing data model
- [x] Unit tests for the newly added functions.
- [ ] Make sure it reproduces simulated missingness

This PR implements part of #43.


seabbs:
So the model now compiles, fits, and works as expected when there is no missing model being used. It looks like it isn't quite there yet as the missingness model is failing to return the simulated proportion missing using the `inst/examples/germany-missing.R` example. This likely indicates a bug with some of the code introduced in this PR. 

I've gone from draft -> actual PR in order to turn on CI but as this still isn't working as expected no review needed (unless feeling very charitable).

seabbs:
I've fixed some of the more problematic bugs with this and added some prototype plotting of the summarised posterior estimates to help with debugging. Still getting slight persistent under-estimation of the proportion of observations missing across a range of simulated scenarios. 

# 30% missing
Screenshot 2022-08-02 at 16 37 23 Screenshot 2022-08-02 at 16 37 54 Screenshot 2022-08-02 at 16 38 07


# 60% missing

Screenshot 2022-08-02 at 16 44 41
Screenshot 2022-08-02 at 16 45 03
Screenshot 2022-08-02 at 16 45 30

seabbs:
No luck tracking this down after doing a read-through of the code. 

Testing using a model where the proportion missing is allowed to vary I see that the bias is located in a given week which might give a hint as to the cause (I assume it is an indexing issue and/or implementation mistake). 

Not to state the obvious but an under-estimate of the proportion missing could mean an over-estimate of the number of people available to be missing or an under-estimate of the people available to be not missing (if that makes sense). 

# Random walk on week

Screenshot 2022-08-02 at 17 24 33

# Random effect on week 

Screenshot 2022-08-02 at 17 32 37

seabbs:
Fitting an even more flexible model with a random effect on the day of the week highlights the issue. The shape of the increase and the initial discontinuity definitely makes me thing there is an index mistake playing a role here. 

Screenshot 2022-08-02 at 17 49 16

inst/stan/functions/delay_lpmf.stan
```diff
@@ -27,33 +27,59 @@ real delay_group_lpmf(array[] int groups, int start, int end, array[] int obs,
...
+    log_exp_complete = expected_obs_from_snaps(
+      i_start, i_end, imp_obs, rdlurd, srdlh, ref_lh, dpmfs, ref_p, rep_h, ref_as_p, sdmax, csdmax, sg, st, f[3]
+    );
+
+    // Allocate to just those actually observed
+    log_exp_obs = allocate_observed_obs(
+      i_start, i_end, log_exp_complete, sl, csl, sdmax, csdmax
+    );
+    log_exp_obs = apply_missing_reference_effects(
+      i_start, i_end, log_exp_obs, sl, csl, log1m(miss_ref_lprop)
```
sbfnk, line 60: Should this not be `log1m_exp`? as `miss_ref_lprop` is already on the log scale?

sbfnk (commented):
The indexing is savage but I couldn't spot any flaws.

inst/stan/functions/delay_lpmf.stan
seabbs: yes it should good spot - it was what causing the initial not working at all issues. I fixed it locally a few commits not at all clear why its showing up not changed here...

seabbs: if I just failed to save the function file that is a lot of time wasted and quite embarrassing!

sbfnk: Would you necessarily expect perfect parameter recovery? There are quite a few small numbers in the toy data set perhaps leading to edge cases given we're using a continuous approximation with a fixed multiplier for missing data. In other words, the simulated data is not from the data generating process of the model - though you could try that (fixed scaling rather than binomial sampling to generate the simulated data set).

sbfnk: moving to main comment section for better threading ettiquette.

sbfnk:
Would you necessarily expect perfect parameter recovery? There are quite a few small numbers in the toy data set perhaps leading to edge cases given we're using a continuous approximation with a fixed multiplier for missing data. In other words, the simulated data is not from the data generating process of the model - though you could try that (fixed scaling rather than binomial sampling to generate the simulated data set).

seabbs:
Yeah maybe not exact but this does look pretty systematic when you dive into it (and the way it is very bad and then iteratively off less bad suggests and off by something index error to me). 

I've tried with a fixed scaling model and the same issue occurs unfortunately.

seabbs:
Did a little more work on this and fixed a small bug in the ordering of data for the simulation that may have been causing issues. 

I have now: 

- Checked on several other snapshots (for example below). Recovery is still not perfect when using a highly flexible modeel but seems much less pathologically incorrect
- Added `enw_incidence_to_cumulative()` to the package and renamed `enw_new_reports()` to match
- Added unit tests for both these functions. 

I still need to: 

- [ ] Update the news item
- [ ] Add unit tests for the internal stan functions where needed
- [ ] Add a regression test for epinowcast that using the missing data model

# Example nowcast (with a daily random walk on the proportion missing)

Screenshot 2022-08-10 at 17 41 27
Screenshot 2022-08-10 at 17 40 58
Screenshot 2022-08-10 at 17 41 12

seabbs:
I've pulled out the new input functions in #157 rather than have them in this PR.

seabbs:
It looks like there was an issue in the R based look-up used to assign reference dates to appropriate report dates (due to using report date to construct the complete data frame of observed data and not reference date). 

See: https://github.com/epiforecasts/epinowcast/pull/147/commits/8e7c1de543fc5122cde8a77270735a7e1c1f6eab

Some more testing needed but early signs look promising. 

Screenshot 2022-08-16 at 19 28 48
Screenshot 2022-08-16 at 19 29 05
Screenshot 2022-08-16 at 19 28 26

seabbs:
Note that in https://github.com/epiforecasts/epinowcast/pull/147/commits/74ad2dffe511f7ccf76cd128ff94230ee50b1098 I have tried to resolve some issues from #151 and carefully checked ordering in all code. This may have also introduced some unintended side effects.

adrian-lison:
Note that I slightly changed the definition of `filt_obs_indexed`, making it more intuitive in my opinion and simplifying the code almost everywhere it is used

inst/stan/functions/filt_obs_indexes.stan
```diff
@@ -1,10 +1,10 @@
...
+// with id `start` and snapshot with id `end`
 array[] int filt_obs_indexes(int start, int end, array[] int csl,
                              array[] int sl) {
   array[3] int n;
-  n[1] = csl[start] - sl[start];
-  n[2] = csl[end];
-  n[3] = n[2] - n[1];
+  n[1] = csl[start] - sl[start] + 1; // index of first obs in snapshot start
+  n[2] = csl[end]; // index of last obs in snapshot end
+  n[3] = n[2] - n[1]; // total number of flattened obs / length of interval
```
seabbs, line 8: Given the updated definition for `n[1]` does this not also need to be updated or was it not correct originally?

adrian-lison: Yes, I also realized that now, sorry!

adrian-lison:
Played around with simulation and preprocessing of missing data and these all look perfectly fine to me.

adrian-lison:
Regarding the comparison in the simulation: because we have no binomial sampling in the missing reference model (shares are deterministic), if we want to be super accurate we should compare the estimated missingness proportions rather to the actually realized missing proprtions than to the binomial probability I think...

seabbs:
Did you look across different dates and proportions of missingness? Have I been having a fever dream 😆

NEWS.md
```diff
@@ -18,10 +21,13 @@ a series of dates. Changed interface of `enw_preprocess_data()` to pass `...` to
 - Introduced two new delay likelihoods `delay_snap_lmpf` and `delay_group_lmpf`. These stratify by either snapshots or groups. This is helpful for some models (such as the missingness module). The ability to choose which function is used has been exposed to the user in `enw_fit_opts()` via the `likelihood_aggregation` argument. Both of these functions rely on a newly added `expected_obs_from_snaps` function which vectorises `expected_obs_from_index`. See #138 by @seabbs and @adrian-lison.
 - Added support for supplying missingness model parameters to the model as well as optional priors and effect estimation. See #138 by @seabbs and @adrian-lison.
 - Refactored model generated quantities to be functional. See #138 by @seabbs and @adrian-lison.
+- Added support for modelling missing reference dates to the likelihood. See #147 by @seabbs.
```
seabbs, line 24: Add Adrian here before merging

adrian-lison:
So I was in a pedantic mood and made some changes to the in-code model documentation to make it a bit more intuitive (at least for me - hoping my brain is not too strange so others would profit as well 😆 )

adrian-lison:
Unfortunately, no solution yet found for the wrong missingsness proportions...

But it really looks like we are leaking observations with missing reference times in the beginning, see the plot of `imp_obs` below (dots are realized / ground truth observations, accounting for the missingness):
imp_obs

Corresponding missingness:
prop_missing

adrian-lison:
This is still a riddle. Using print statements in stan, and copying the results into R (as a replacement for debugging), I have verified that no observations from imp_obs get lost further downstream:

- [x] `imp_obs` get fully distributed across delays
- [x] correct shares of cases with missing and non-missing reference allocated, sum up to overall expected obs
- [x] obs with missing reference correctly allocated to reporting times

inst/stan/functions/delay_lpmf.stan
```diff
@@ -27,33 +27,59 @@ real delay_group_lpmf(array[] int groups, int start, int end, array[] int obs,
...
+    log_exp_obs = expected_obs_from_snaps(
+      i_start, i_end, imp_obs, rdlurd, srdlh, ref_lh, dpmfs, ref_p, rep_h, ref_as_p, sl, csl, sg, st, n[3]
+    );
+  }
   // Observation error model (across all reference times and groups)
   profile("model_likelihood_neg_binomial") {
   tar = obs_lpmf(filt_obs | log_exp_obs, phi, model_obs);
+  if (model_miss && miss_obs) {
+    tar += obs_lpmf(missing_reference | log_exp_obs_miss, phi, model_obs);
+  }
```
adrian-lison, line 82: @seabbs Are we maybe still missing the filtering for the current group here? I.e. the parallel to 

```
array[3] int n = filt_obs_indexes(i_start, i_end, csl, sl);
array[n[3]] int filt_obs = segment(obs, n[1], n[3]);
```
for `missing_reference`?

adrian-lison:
Since I am running out of possible implementation-level explanations for the observed behavior, I am starting to think about the model specification...

What seems to happen (not sure if this is the only problem) is that if specified flexibly, the share of cases with missing reference can compensate for a bad fit of `imp_obs` by producing too many cases with missing reference and distributing them over the delays. Since the likelihood for the cases with missing reference is less sharp (integrating over delays) the model can get away with that especially if case numbers are low, see the examples below.

My question is whether this is maybe an artefact of the parametric delay distribution not fitting super well to this real-world dataset. Perhaps we should do some tests with simulated data where everything is well-behaved.
I could also feed the Germany dataset with the simulated missingness into my model implementation (since it has a non-parametric reporting delay) and see what happens.

### Example 1  (40% missingness)
02 imp_obs long
02 miss prop long
(note that the dashed line is the realized missing proportion through binomial sampling in the simulation)

It even looks as if the share of cases with missing reference is also capturing some of the weekday effect... This would be quite weird, and I haven't seen something like that with my models.

### Example 2 (10% missingness)
03 imp_obs
03 miss prop
This basically illustrates that the problem does not go away even if missingness is small.

seabbs (commented):
@adrian-lison changes look good. Updates to model documentation are a big improvement.

seabbs:
> Unfortunately, no solution yet found for the wrong missingsness proportions...

I thought that this issue was maybe due to how we define the random walk (i.e it starts from `t = 0` where data is sparse) vs the issue found elsewhere? As we discussed it's not totally clear to me what to do about this with the current implementation and it perhaps points to a flaw that should be corrected.

seabbs:
> Since I am running out of possible implementation-level explanations for the observed behaviour, I am starting to think about the model specification...

The model being overly flexible in the wrong places is a sensible suggestion. If we moved testing to the implementations that also have a flexible expectation model that would perhaps help explore this? 

> the model can get away with that especially if case numbers are low, see the examples below.

Could this also be a tension between picking a more extreme growth rate and the right missingness proportion? The log-normal random walk may limit the amount of change over time required to fit this data well.

> My question is whether this is maybe an artefact of the parametric delay distribution not fitting super well to this real-world dataset. Perhaps we should do some tests with simulated data where everything is well-behaved.
I could also feed the Germany dataset with the simulated missingness into my model implementation (since it has a non-parametric reporting delay) and see what happens.

This is a good suggestion and could well be the root cause. Both options for exploring are sensible. In general, we are crying out for a model simulator to ease testing and boost confidence. 

Something that also concerns me is that perhaps this is highlighting a bug elsewhere. Perhaps with model specification or allocation of delays? I was fairly confident that was all working as intended but it hasn't recently been tested on clean simulated data (i.e not since initial development). 

> It even looks as if the share of cases with missing reference is also capturing some of the weekday effect... This would be quite weird, and I haven't seen something like that with my models.

This I found quite concerning and enforces my concern there is a bug elsewhere. Seems quite hard to explain/understand.

In terms of moving this forward, I wonder if perhaps we should merge this into `develop` with a big warning pointing back at this PR. Then work up #152 and #154 to MVP status and merge, and after that is done circle back to this before release. More eyes on this would likely help and it's starting to get quite weighty.

(this is a great investigation by the way, can you put the exploratory code for this somewhere so we can explore more in the future?)

inst/stan/functions/delay_lpmf.stan
seabbs: ah, yes I think you are right. I don't see how this can be correct as written when there are multiple groups present.

seabbs: Note this shouldn't be linked to the issues we see as all testing so far done on only a single group.

seabbs:
Proposed tasks to get an MVP to merge: 

- [ ] Fix `missing_reference` where multiple groups are present
- [ ] Add experimental warning
- [ ] Add Adrian to news items

inst/stan/functions/delay_lpmf.stan
adrian-lison: Yeah, agree, this is a different topic

adrian-lison:
Hey sam, thanks for your replies! I agree that adding this to develop may be a good idea. I think I can work on the open tasks as suggested above by you.

Can add my investigation code to the example script we already have.

For my own models, I have written a simulator that creates linelists based on a renewal process, which can then be used to create nowcasting datasets. I could maybe create a PR for this, although it would be more work to get it into the style of epinowcast (need to switch from dplyr to data.table e.g.).

NEWS.md
seabbs: Resolved

inst/stan/functions/delay_lpmf.stan
seabbs: I've resolved this in https://github.com/epinowcast/epinowcast/pull/147/commits/2ec8058155fa4b8f6ddc2f3833803b3d90ef93e9

seabbs:
So I think I have covered the minimum ground required to merge this as is..... 

> Can add my investigation code to the example script we already have.

I think this would be useful. 

> For my own models, I have written a simulator that creates linelists based on a renewal process, which can then be used to create nowcasting datasets. I could maybe create a PR for this, although it would be more work to get it into the style of epinowcast (need to switch from dplyr to data.table e.g.).

Yeah I think having something in the repository we can play with would be great - regardless of implementation style. Perhaps lets add to the repo and put in the `.Rbuildignore` to avoid having to add as a suggests?

seabbs:
As discussed I have gone ahead and merged this into `develop` as an experimental feature.

Issue #144, opened 2022-07-27
Add data format converters


seabbs:
At the moment we require our input data as a cumulative count. This makes sense for many classic nowcasting problems but makes less sense for some forms of data and in particular linelist data.

We should consider adding the following new functions. 

- [x] `enw_linelist_to_incidence()` aggregate a individual linelist to aggregated counts
- [x] `enw_incidence_to_cumulative()` aggregate incidence count data to cumulative.
- [x] `enw_cumulative_to_incidence()`
- [x] `enw_incidence_to_linelist()`

These functions are also likely useful for exploring data and potentially simulation. We want them to support `by` arguments as discussed in #143.

There is a version of `enw_incidence_to_cumulative()` as part of #138 which can be pulled out and improved. 

`enw_new_reports()` is effectively `enw_cumulative_to_incidence()`. It could be renamed and depreciated using `lifecycle`.

seabbs:
@sbfnk this is likely good functionality for a spin-off/another package. Parts of this are available in, for example, `incidence2` but I don't think on a way that is sufficiently generic/flexible for our use case.

seabbs:
I made a start on `enw_incidence_to_cumulative()` and renamed `enw_new_reports()` to `enw_cumulative_to_incidence()` as part of trying to debug simulation code for the missing data model.

seabbs:
I've pulled out work on `enw_incidence_to_cumulative()` and `enw_cumulative_to_incidence()` in #157.

jhellewell14:
One thing to choose I think is what exactly you want out. At the moment `cumul_to_inc` adds a column `new_confirm` and the column `confirm` continues to exist so that data.table now has both cumul and inc. With linelist conversion what you get out looks nothing like the original linelist. 

So maybe it would be better to think of it as three functions:
1) `enw_add_inc_to_cumul` (adds `new_confirm` and % reported cumulatively if not present to incidence data)
2) `enw_add_cumul_to_inc` (adds `confirm` and % reported cumulatively to cumulative data)
3) `enw_linelist_to_dt` (produces dt with `confirm`, `new_confirm`, and % reported from linelist data)

jhellewell14:
I've put a start at the linelist conversion function here: https://github.com/jhellewell14/epinowcast/commit/65463ddb0541a1f2fe676224b3daeb125b01e086 

I have also included a fake linelist dataset that might be relevant for part of https://github.com/epiforecasts/epinowcast/issues/94

[...]

PR #157, opened 2022-08-11
Feature incidence to cumulative

seabbs:
This PR pulls out changes from #147 to `enw_new_reports()` and adds `enw_incidence_to_cumulative()` also developed in that PR. 

It represents a prototype for a subset of the tasks described in #144 but these may not be final versions. @jhellewell14 any thoughts appreciated. 

To do:

- [x] News item


seabbs (commented):
LGTM (self-review)

PR #160, opened 2022-08-19
Rewrite some formula_tools functions without regex

Bisaloo:
Putting this here for discussion. Not sure if it's desirable or not.

Only change compared with current version is that `remove_rw_terms()` is now associated with an environment. It can probably be changed but I'm not completely sure why formulas are usually linked to environments so I didn't touch it.


seabbs:
Thanks for this Hugo. The use of terms looks nice. Currently seeing some issues with this in the tests that makes me think needs a tweak. 

Also I note there has been a slightly functionality change (the random effects handling) that needs telegraphing to users.

seabbs:
I'm going to close this for now but happy to reopen and discuss.

Issue #156, opened 2022-08-08
Automatic detection of `choices` in `match.arg()`


Bisaloo:
The base function `match.arg()` is used throughout the code to check that values provided for arguments are valid input. Each time, the `choice` argument to indicate what values are valid. Some examples:

https://github.com/epiforecasts/epinowcast/blob/603782bece7cea70d68287d4481fc0dc08a006b8/R/formula-tools.R#L250

https://github.com/epiforecasts/epinowcast/blob/603782bece7cea70d68287d4481fc0dc08a006b8/R/methods.R#L23-L26

This is not strictly necessary since R already includes a mechanism to automatically detect the value of this choice argument:

``` r
f <- function(arg1 = c("value1", "value2")) {
  
  # This line is equivalent to:
  # arg1 <- match.arg(arg1, choices = c("value1", "value2"))
  arg1 <- match.arg(arg1)
  
  return(arg1)
  
}

f()
#> [1] "value1"
f("value1")
#> [1] "value1"
f("value2")
#> [1] "value2"
f("wronginput")
#> Error in match.arg(arg1): 'arg' should be one of "value1", "value2"
```

<sup>Created on 2022-08-08 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1.9000)</sup>

**Should `choices` be provided in `match.arg()` calls in epinowcast or should it rely on the automatic detection feature?**

Some pros/cons discussed with @seabbs this morning:

- Current style:
  * the list of choices is visible on the line where `match.arg()` is called rather than possibly having to scroll up to the function definition
  * slightly faster
 
- Proposed style:
  * the list of choices is visible in the documentation (although it is good practice to re-document it in the relevant roxygen2 `@param` section)
  * slightly more idiomatic. This is used throughout base R code as well as commonly used packages such as the tidyverse.

Bisaloo:
To weigh in, I don't think the two arguments in favour of the current style are very strong:

> the list of choices is visible on the line where match.arg() is called rather than possibly having to scroll up to the function definition

`match.arg()`, as every other input checking, should almost always be called right at the start of the function.

> slightly faster

It is faster in benchmarks but it is not likely to be significant compared to the rest of the function. This step will always be fast, no matter what style is chosen.


Overall, I would lean towards the more idiomatic option.

seabbs:
Thanks again Hugo and thanks for making such a clear issue (perhaps we should link to this from the contributing guide as a good example). 

[...]

PR #162, opened 2022-08-26
address match.args change

pearsonca:
Addresses #156, with the exception of some esoteric `match.arg` uses


pearsonca:
I can update that part. Is there also place more broadly to document the preferred use of `match.arg`? Something style-guide ish?

seabbs:
I think perhaps the part that says following the style of the code may cover it? We don't have a very detailed style document yet where this would live (probably scare people with detail if it went in the more general one we do have).

pearsonca:
alright; i think this too specific for the current level of discussion in CONTRIBUTING, so i've just noted in the NEWS write up.

pearsonca:
aside: bit surprised that tweak to NEWS file only triggers all the check actions. anyway to exclude changes to certain files (e.g. NEWS, CONTRIBUTING, LICENSE come to mind) from triggering that?

seabbs (approved):
LGTM. ❤️

PR #169, opened 2022-11-30
Feature: interactions in random effects

seabbs:
This PR adds support for interactions on the right-hand side of random effects (i.e `(1 + month | day_of_week:age_group)`. This should result in a model with an independent random effect for each age group on the day of the week and random slopes on the month drawn from separate hyper distributions for each age group.

This feature is motivated by #152 where we need to be able to specify independent random effects for each group easily in order to reproduce the current default model. There may also be other use cases in the future. 

Potentially this syntax is not ideal and/or there are other work around to this feature that I am not aware of. 

I have added: 

- Functionality to support random effect interactions
- An example in the documentation of `enw_formula` showcasing this functionality
- A test for `enw_formula` that produces a snapshot. I have confirmed this snapshot has the desired design matrix structure. 
- A test checking that multiple interactions fail with a meaningful error (needed as this is more complex to implement and so is not yet in place).
- A news item highlighting this change.

I have also checked that none of the other functionality has been impacted by reviewing our other tests.

To test out the new functionality see below: 

```r
obs <- enw_filter_report_dates(
  germany_covid19_hosp[location == "DE"][
    age_group %in% c("00+", "05-14", "15-34")
  ],
  remove_days = 10
)
obs <- enw_filter_reference_dates(obs, include_days = 10)
pobs <- enw_preprocess_data(obs, by = c("age_group", "location"))
data <- pobs$metareference[[1]]

enw_formula(~ 1 + (1 + month | day_of_week:age_group), data)
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -16,6 +16,7 @@ a series of dates. Changed interface of `enw_preprocess_data()` to pass `...` to
 - Skipped tests that use `cmdstan` locally to improve the developer/contributor experience. See #147 by @seabbs and @adrian-lison.
 - Added a basic simulator function for missing reference data. See #147 by @seabbs and @adrian-lison.
 - Added a basic simulator function for missing reference data. See #147 by @seabbs and @adrian-lison.
+- Added support for right hand side interactions as syntax sugar for random effects. This allows the specification of, for example, independent random effects by day for each strata of another variable. See #169 by @seabbs.
 
 ## Model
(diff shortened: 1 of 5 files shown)


seabbs:
@epinowcast/developers if anyone has time to review this that would be appreciated. There will be another formula based PR coming shortly to support optional intercepts needed by #152.

seabbs (commented):
LGTM (self-review)

Issue #179, opened 2022-12-05
Separate latent reporting delays from latent reporting scaling


seabbs:
In #152 a latent reporting process was implemented that supported both delays in reporting and partial ascertainment by allowing the specification of a PMF that does not sum to 1. As @adrian-lison notes it may make more sense to separate this functionality into two arguments which could be called `latent_reporting_fraction` and `latent_reporting_delay`. Not that any changes along these lines need to be made with #177 in mind and this will substantially complicate both the front and back-end implementation.

seabbs:
In #183 I've thought about it some more and based on needing time-varying reporting delays (i.e #177) and potentially group-specific ones it feels like keeping this as a linked option, for now, maybe simpler. If we do split out then we will need to manage the various permutations of both arguments and then combine them. 

Once we have better control of priors then manipulating the `observation` formula becomes a better option (though perhaps this also becomes complex when/if moving to a hazard model for this. 

Given this going to propose this issue gets closed out. What do you think @adrian-lison?

adrian-lison:
Yes, I agee, with #183 in place now it would rather be a pain to separate the two, and not improve readibility. So let's close this.

Issue #174, opened 2022-12-05
Test coverage for renewal and latent reporting process


seabbs:
#152 added both a renewal process and a latent reporting process. Neither of these currently has any test coverage. To some degree this depends and is linked to #173 as the current implementation is so unstable it is hard to test. However, even with this implementation stan level tests are possible.

seabbs:
This is currently blocked as `epinowcast:::expose_stan_fns` cannot be used to expose these functions for testing in R. This is because `convert_cmdstan_to_rstan` cannot currently produce stan code that is valid in `rstan` due to the use of array vectors. For this reason, I have bumped this issue from our current release timelines.

seabbs:
#192 from @sbfnk enables unit tests for `log_expected_obs_from_latent.stan`

However `expose_stan_fns("log_expected_latent_from_r.stan", "inst/stan/functions")` due to issues wit vectoor assignment. This likely also requires a change in syntax processing.

Issue #173, opened 2022-12-05
Log -> Real renewal and latent reporting


seabbs:
In #152 a log scale renewal process and latent reporting process were added. These have been found to be unstable and slow to fit. An obvious alternative is to fall back to the usual approach of modelling on the real scale with a dot product. Some exploration of this has been started in #168 and on a first pass, it does appear to increase stability and reduce run times.

seabbs:
Some fairly confusing results on this: https://github.com/epinowcast/epinowcast/pull/183

seabbs:
#183 has been cleaned up and extended. Tests showed improved performance and moderate increases in stability with no impact on results.

Issue #172, opened 2022-12-05
Document the flexible expectation model and optional renewal process + latent reporting.


seabbs:
#152 added a flexible expectation model (where the expectation process can now be specified using a formula interface vs being hardcoded as a lognormal random walk). This needs to be documented in `vignettes/model.Rmd` as well as being clear as an option in other introductory documentation.

#152 also added an optional renewal process and latent reporting process. Both of these are currently undocumented with the documentation needed in the same places as for the flexible expectation process.

As noted in our development guidelines any changes should be targeted at the `develop` branch.

seabbs:
I've started fleshing this out in #189

seabbs:
Done in #189

PR #180, opened 2022-12-05
0.2.0 Release

seabbs:
This PR is for the `0.2.0 epinowcast` release.  

Outstanding work to do before releasing to main:

- [ ] #173 
- [ ] #172 
- [ ] #167
- [ ] #166 
- [ ] #174 
- [ ] #179
- [ ] Add links to the community.epinowcast.org site
- [ ] Add links to the example scripts for various models (i.e in `inst/examples/*.R`)
- [ ] Add a summary section to the release note and thank all authors


adrian-lison:
Thanks a lot for compiling this, @seabbs! I agree with the outline, and would only potentially add #176 and #175 to the list. This is because our latent renewal model component is still a bit experimental, and if we want to supply this to users, it may be valuable to give them some basic debugging capabilities.

Or are you thinking the other way round, that we shouldn't make the R / growth rate estimation too prominent for now, to avoid that rather novice users treat the results without caution?

seabbs:
> Or are you thinking the other way round, that we shouldn't make the R / growth rate estimation too prominent for now, to avoid that rather novice users treat the results without caution?

Yes, this was my thought and also that I find this kind of functionality a bit tricky to make sufficiently general. I was concerned we would implement something and then to iterate. If we instead give code examples for manual extraction/plotting etc perhaps we can find the right design a bit more organically? 

I definitely agree both of these things should be a priority for the next release though.

What do you think?

seabbs:
All release blockers have been resolved. Moving this to `main` as a candidate release.

PR #182, opened 2022-12-06
Feature cpp options

seabbs:
This PR adds support for passing `cpp_options` this is useful for example when using GPU support (testing this out indicates it doesn't help on small scale models and in fact leads to a large slow down).


seabbs (commented):
LGTM (self-review)

PR #183, opened 2022-12-06
Renewal and latent reporting on the real scale

seabbs:
As flagged in #173 there is more evidence that a natural scale-based approach using `dot_products` etc is likely to be more stable and faster to fit than the current log scale implementation. I explore this in this PR using two minimal refactors to go from the log scale to the natural scale. 

Using `inst/examples/germany_latent_renewal.R` as a test bed. I fit the following models over 20 chains for the original log scale implementation, a real scale renewal process, and a log scale latent reporting process, and both processes on the real scale. I found the following:

- Daily random walk on the log of expected case
     - Log: 83 seconds with one divergent transition
     - Log latent and real renewal: 78 seconds
     - Real: 79 seconds

- Weekly random walk on the growth rate/Rt with adapt delta set to 0.9
     - Log: 141 seconds and 2% DTs
     - Log latent and real renewal: 132 seconds and 4% DTs
     - Real: 137 seconds and 4% DTs

Given these results, I am a little unsure how to proceed. Note that fairly directly contradict @adrian-lison investigations that found a large performance regression using the log scale approach. Potentially an issue is that the implementations are not optimised and if more is done on the real scale vs the log scale then the performance benefits will become more obvious. I thought about matrix implementations (to replace the vectorised for loops) but it is not clear to me this will lead to a large speed-up as the matrices will need to be allocated in a loop regardless of the calculation that then takes place.


seabbs:
So I just extended the testing to 500 chains and found that perhaps the main difference in this instance is for edge cases. The log scale approach had multiple chains that had very long runtimes whereas the natural scale approaches didn't (though I did see a few chain failures in both which isn't ideal). 

This is obviously wrapped up with other issues (i.e we need to make initialisation better) but makes a clearer case than the smaller tests for a real scale implementation.

seabbs:
I had a dream that matrix multiplication was a good idea (based on a desperate need to use a GPU). It isn't totally clear if the dream was real.

On top of the matrix multiplication, I also implemented a very rough (i.e this is not final so don't fret) sparse matrix version. This should in theory be the best of both worlds as it deals with the issues of lots of zeros in the matrix whilst still allowing for a hopefully more optimised model. Unfortunately, it is quite opaque. 

The side benefit of doing it this way is we can easily support time-varying reporting delays as long as the user is somewhat dedicated to mucking around with the code (we can obviously improve going forward). Even if we roll back to the `dot_product` version then keeping some input matrix structure is likely a good idea so this continues to be the case.

I tested using `germany_latant_renewal.R` and the settings seen in the PR. Across 50 chains both implementations gaves similar runtimes to the `dot_product` based approach but the pure matrix method had two failing chains.

Code

- https://github.com/epinowcast/epinowcast/pull/183/commits/7d78d3a923e5f18088edfce829f5019eedf11f90 matrix based approach
- https://github.com/epinowcast/epinowcast/pull/183/commits/7afacea8fe55442eb3538f358254509730da6d87 sparse matrix-based approach

seabbs:
I've extended the current approach to add time-varying reporting distributions as well as adding a helper function to allow the combination of PMFs. I've added this to the news and will update the example etc with a doc overhaul in another PR.

seabbs (commented):
LGTM (self-review). Opinion based on read-through of code and tests mentioned in the PR.

seabbs:
Going to merge this for now into `develop` but very happy to see critique either here or in a new issue.

PR #185, opened 2022-12-11
Add target dir passing to enw_model

seabbs:
This PR adds the ability to pass the `target_dir` argument from `write_stan_files_no_profile` from `enw_model`. This is helpful when using the model in multiple R sessions and wanting to share a single compiled file.


seabbs (commented):
LGTM (Self-review). I tested locally and this has no impact for default users but works as advertised when specified. Care needed not to overwrite the package model and cause confusion.

PR #187, opened 2022-12-12
Redo add_pmf, fixes #186

### LINKED ISSUE #186: Optimise add_pmfs
https://github.com/epinowcast/epinowcast/blob/1b39e77447d0b858afa810806b8984f0945f009b/R/model-module-helpers.R#L173

core replacment fits in a tweet:

```
Reduce(\(conv, pmf) {
    lc <- length(conv)
    wd <- (1:lc)-1
    proc <- numeric(lc + length(pmf))
    for (j in seq_along(pmf)) {
      proc[j+wd] <- proc[j+wd] + pmf[j] * conv
    }
    proc
  }, pmfs)
```

If this is actually hot in the performance chain (doubt it, but ... :shrug:), then can just replace that with some cpp.

@pratikunterwegs

pratikunterwegs:
Modifies `add_pmfs` per @pearsonca, fixes #186

### DIFF
--- a/R/model-module-helpers.R
+++ b/R/model-module-helpers.R
@@ -165,7 +165,7 @@ convolution_matrix <- function(dist, t, include_partial = FALSE) {
 #' # Analytical convolution of PMFs
 #' conv_pmf <- add_pmfs(list(xpmf, ypmf))
 #' conv_cdf <- cumsum(conv_pmf)
-#'# Empirical convolution of PMFs
+#' # Empirical convolution of PMFs
 #' cdf <- ecdf(z)(0:42)
 #' # Compare sampled and analytical CDFs
 #' plot(conv_cdf)
@@ -178,23 +178,16 @@ add_pmfs <- function(pmfs) {
   if (!is.list(pmfs)) {
     return(pmfs)
   }
-  lpmfs <- purrr::map_dbl(pmfs, length)
-  l <- sum(lpmfs)
-  conv <- rep(0, l)
-  conv[1:lpmfs[1]] <- pmfs[[1]]
-  for (s in 2:d) {
-    # P(Z = z) = sum_over_x(P(X = x) * P(Y = z - x)) # nolint
-    proc <- rep(0, l)
-    for (i in 1:l) {
-      for (j in 1:lpmfs[s]) {
-        if (i >= j) {
-          proc[i] <- proc[i] + pmfs[[s]][j] * conv[i - j + 1]
-        }
-      }
+  conv <- Reduce(x = pmfs, f = function(conv, pmf) {
+    lc <- length(conv)
+    wd <- seq_len(lc) - 1
+    proc <- numeric(lc + length(pmf))
+    for (j in seq_along(pmf)) {
+      proc[j + wd] <- proc[j + wd] + pmf[j] * conv
     }
-    conv <- proc
-  }
-  return(conv)
+    proc
+  })
+  conv
 }
 
 #' Extract sparse matrix elements


pratikunterwegs:
Can't seem to run CI workflows myself, but passes tests and builds locally.

R/model-module-helpers.R
```diff
@@ -178,23 +178,16 @@ add_pmfs <- function(pmfs) {
...
-  for (s in 2:d) {
-    # P(Z = z) = sum_over_x(P(X = x) * P(Y = z - x)) # nolint
-    proc <- rep(0, l)
-    for (i in 1:l) {
-      for (j in 1:lpmfs[s]) {
-        if (i >= j) {
-          proc[i] <- proc[i] + pmfs[[s]][j] * conv[i - j + 1]
-        }
-      }
+  conv <- Reduce(x = pmfs, f = function(conv, pmf) {
```
pearsonca, line 181: Is the assignment / last line thing indicated by the styler? Seems weird, but I'm terse.

I prefer the bare "last evaluation" return approach (though not as dev friendly); if not that, wrap in `return(...)` (more dev friendly, marginal performance hit). Assignment is an unnecessary symbol allocation / parsing step.

pratikunterwegs: It isn't - that's just my concession to readability - happy to wrap the `Reduce` in a `return` if that's more the package style

seabbs: I'd prefer to see returns used to match package style.

seabbs:
Thanks for this @pratikunterwegs and @pearsonca. 

I've just enabled CI checks and will let those tick over before reviewing. @pratikunterwegs there are a couple more bits it would be nice to see here before can merge like an addition to the news file and yourself as a contributor (if you would like).

Contributing guide: https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md

pratikunterwegs:
> Thanks for this @pratikunterwegs and @pearsonca.
> 
> I've just enabled CI checks and will let those tick over before reviewing. @pratikunterwegs there are a couple more bits it would be nice to see here before can merge like an addition to the news file and yourself as a contributor (if you would like).
> 
> Contributing guide: https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md

Thanks! It's @pearsonca's work so wouldn't add myself as contrib. I looked over the news file - would you prefer I start a section for this minor version (0.2.0), or a new one, or for this dev version (0.2.0.1000) - happy to go with the package convention. Also some checks mentioned in the contributing guide (e.g. `goodpractice::gp()`, linting) throw quibbles in files not targeted in this PR - I'm going to assume you prefer that I don't modify those but say otherwise if so.

R/model-module-helpers.R
pratikunterwegs: Fixed in https://github.com/epinowcast/epinowcast/pull/187/commits/cc5ecb4e92671dc81d46544f923c38cc24ff6b4a

seabbs: The internal `return` is still missing and to be honest, I would prefer the slight performance hint of having an assignment followed by a return. I'm happy to read links that argue against that as a stylistic choice of course.

R/model-module-helpers.R
```diff
@@ -178,23 +178,17 @@ add_pmfs <- function(pmfs) {
...
-    for (i in 1:l) {
-      for (j in 1:lpmfs[s]) {
-        if (i >= j) {
-          proc[i] <- proc[i] + pmfs[[s]][j] * conv[i - j + 1]
-        }
+  return(
+    Reduce(x = pmfs, f = function(conv, pmf) {
+      lc <- length(conv)
+      wd <- seq_len(lc) - 1
+      proc <- numeric(lc + length(pmf))
```
seabbs, line 185: as this drops the comment equation I think additional documentation may be needed. I can add this before merging or happy to see it added.

R/model-module-helpers.R
```diff
@@ -178,23 +178,17 @@ add_pmfs <- function(pmfs) {
...
+      for (j in seq_along(pmf)) {
+        proc[j + wd] <- proc[j + wd] + pmf[j] * conv
       }
-    }
-    conv <- proc
-  }
-  return(conv)
+      proc
```
seabbs, line 189: As discussed I would rather see `return` used explicitly here.

R/model-module-helpers.R
```diff
@@ -178,23 +178,17 @@ add_pmfs <- function(pmfs) {
...
-  conv[1:lpmfs[1]] <- pmfs[[1]]
-  for (s in 2:d) {
-    # P(Z = z) = sum_over_x(P(X = x) * P(Y = z - x)) # nolint
-    proc <- rep(0, l)
-    for (i in 1:l) {
-      for (j in 1:lpmfs[s]) {
-        if (i >= j) {
-          proc[i] <- proc[i] + pmfs[[s]][j] * conv[i - j + 1]
-        }
+  return(
```
seabbs, line 181: This is a stylistic choice and one I wouldn't usually take. Happy to leave as is after reflection.

seabbs (changes requested):
Thanks this looks good. Just the pedantic point and the missing return and a note about potentially needing to extend docs to maintain interpretability.

seabbs:
This looks good I think. I've noted the pedantic issues about the use of `return`.

I'm happy to see you added for this PR (I would also like to add the more permissive all contributors bot which helps with this but yet to have a chance to implement 😉 ). 

> section for this minor version (0.2.0), 

On the `develop` branch you should see an in progress news entry for `0.2.0`. I think this would fit in the package subsection. 


> throw quibbles in files not targeted in this PR - I'm going to assume you prefer that I don't modify those but say otherwise if so

Yup leave here but agree there is some clean up to do (which I am happy to see in a PR!)

R/model-module-helpers.R
pratikunterwegs: Would you prefer assigment to `conv` and then `return()`?

R/model-module-helpers.R
pratikunterwegs: Would you prefer the commented equation re-added, a plain text description, or would you prefer to add a more concise equation? The last I'd leave to you to be sure it's correct.

pratikunterwegs:
> On the `develop` branch you should see an in progress news entry for `0.2.0`. I think this would fit in the package subsection.

Added following style in previous entries, in https://github.com/epinowcast/epinowcast/pull/187/commits/f32cfaaf2dcf9209866ae5ed9c8fb78cf605481d; please see for style.

R/model-module-helpers.R
pratikunterwegs: Fixed in https://github.com/epinowcast/epinowcast/pull/187/commits/bdf172ebeff5eb4b3882cdab6f0e4a3d8abe228f; please check if it follows style.

R/model-module-helpers.R
pratikunterwegs: Commented equation re-added in https://github.com/epinowcast/epinowcast/pull/187/commits/bdf172ebeff5eb4b3882cdab6f0e4a3d8abe228f.

R/model-module-helpers.R
pratikunterwegs: Changed to explicit `return` but no assignment in https://github.com/epinowcast/epinowcast/pull/187/commits/bdf172ebeff5eb4b3882cdab6f0e4a3d8abe228f

R/model-module-helpers.R
seabbs: Looks good. Thanks.

R/model-module-helpers.R
seabbs: yup looks good. I think we could flesh out the documentation but happy to leave that for a future PR etc just having the commented equation back in is enough I think to hand hold users reading the code.

seabbs (approved):
This looks good. Just waiting for checks and then will merge. Thanks for implementing @pratikunterwegs (and welcome to the `epinowcast` party) and thanks @pearsonca for the suggested refactor and review.

PR #189, opened 2022-12-14
Add renewal and latent reporting to model overview

seabbs:
This work in progress PR adds the renewal equation generative process and the latent reporting convolution to the model overview vignette.


seabbs (commented):
LGTM (self-review)

PR #190, opened 2022-12-16
Added a warning about empirical and specified delay being different

### LINKED ISSUE #167: Error when the specified maximum delay is greater than that observed in the data
Currently when the maximum delay is greater than that observed in the data an error occurs. In reality this should really lead to a warning/message and the maximum delay should be updated to be the maximum observed in the data.

Alternatively when wanting to not use a truncated distribution the default could be to make the code robust to this (though likely with a warning to prevent naive users having very long run-times).

seabbs:
This PR closes #167 by adding a warning when the maximum empirical delay is shorter than the specified maximum delay. It also adds unit tests for `enw_delay_filter` and adds a unit test to check that `epinowcast` fits as expected using its default settings (this catches a bug in the settings for `enw_fit_opts` that was caught whilst developing this PR.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -80,5 +80,5 @@ Encoding: UTF-8
 Language: en-GB
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.2
+RoxygenNote: 7.2.3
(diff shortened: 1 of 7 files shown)


seabbs (commented):
LGTM (self-review)

PR #192, opened 2022-12-19
compatibility with additional array notation

sbfnk:
Updates `convert_cmdstan_to_rstan` to deal with some previously unsupported array notation. 

This may address the issue mentioned in https://github.com/epinowcast/epinowcast/issues/174#issuecomment-1356889203.


seabbs (approved):
Nice (though I really wish we didn't need to do this and it makes my brain hurt). 

This now allows us to expose `log_expected_obs_from_latent.stan` (i.e using `expose_stan_fns("log_expected_obs_from_latent.stan", "inst/stan/fun
    ctions")`).

PR #193, opened 2023-01-05
Featur - update-documentation

seabbs:
This PR updates the package vignettee to showcase the new flexible expectation model. It also fixes a small bug in the expectation model that impacted hierarchical models only. In additon to updating the vignette the README has also been updated


seabbs (commented):
There is some confusing periodicity present in the posterior prediction plot that was not present the last time this vignette was updated. I also see some changes in relative performance that need to be checked. Checking these interactively.

Otherwise, everything looks good.

seabbs:
Checking other models it appears the issue is widespread when using multiple groups. This could indicate an issue with updates to the stan code or it could indicate a new issue with preprocessing.

seabbs:
https://github.com/epinowcast/epinowcast/pull/193/commits/23ccecd6ce7d8b3b37f46b4bacbcba964448ac4c catches a small bug that has been introduced into `enw_pp_summary()` and not caught previously. This corrects the issue with the posterior prediction plots.

seabbs:
This now looks good to me.

Issue #198, opened 2023-02-28
Tests failing for enw_score_nowcast


seabbs:
Tests are currently failing for `enw_score_nowcast` which is a thin wrapper for `scoringutils::score()` to enable ease of use. Testing suggests the failure is coming from with the `score()` function and may be related to the recent `data.table` CRAN release. 

It is unclear if this issue is on the `scoringutils` side or if our usage is leading to an undocumented failure state. Pinging @nikosbosse for info (this may need a separate issue on the `scoringutils` side if changes are needed there.

This issue impacts an optional feature and so is not likely to critically impact users of the package. 

You can reproduce this issue with the following:

``` r
library(epinowcast)

nowcast <- enw_example("nowcast")
summarised_nowcast <- summary(nowcast)
obs <- enw_example("observations")

enw_score_nowcast(summarised_nowcast, obs)
#> Loading required namespace: scoringutils
#> The following messages were produced when checking inputs:
#> 1.  There is no column called `model` in the data. scoringutils therefore thinks that all forecasts come from the same model
#> Error in vec_arith.pillar_num.default("*", e1, e2): is.numeric(y) is not TRUE
```

<sup>Created on 2023-02-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">

<summary>Session info</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.2.2 (2022-10-31)
#>  os       Ubuntu 22.04.2 LTS
#>  system   x86_64, linux-gnu
#>  ui       X11
#>  language en_GB:en
#>  collate  en_GB.UTF-8
#>  ctype    en_GB.UTF-8
#>  tz       Europe/London
#>  date     2023-02-28
#>  pandoc   2.9.2.1 @ /usr/bin/ (via rmarkdown)
#>
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package        * version    date (UTC) lib source
#>  abind            1.4-5      2016-07-21 [1] CRAN (R 4.2.2)
#>  backports        1.4.1      2021-12-13 [1] CRAN (R 4.2.2)
#>  boot             1.3-28     2021-05-03 [2] CRAN (R 4.2.2)
#>  callr            3.7.3      2022-11-02 [1] CRAN (R 4.2.2)
[...]

PR #200, opened 2023-02-28
Seabbs/issue199

### LINKED ISSUE #199: Support for {touchstone} based benchmarking.
`{touchstone}` is an R package for benchmarking functionality in a branch against the PR target. We make use of it in [`{EpiNow2}`](https://github.com/epiforecasts/EpiNow2/tree/main/touchstone)  (mainly implemented by @sbfnk) were it allows us to make sure we don't accidently introduce performance regressions. Its already proved useful in identifying [issues with new features](https://github.com/epiforecasts/EpiNow2/pull/363#issuecomment-1428427959). 

It would be great to have in `{epinowcast}`. In the first instance wrapping it around the examples in `inst/examples` seems like it could be a good idea?

seabbs:
This PR closes #199 by adding `{touchstone}` benchmarks using the package script examples. See #238 for a test implementation.


test.R
```diff
@@ -0,0 +1,7 @@
+library(epinowcast)
+
+ nowcast <- enw_example("nowcast")
+ summarised_nowcast <- summary(nowcast)
+ obs <- enw_example("observations")
+
+enw_score_nowcast(summarised_nowcast, obs)
```
adrian-lison, line 7: What is this script for?

seabbs: ack - that is left over from making a reprex for #198 will remove.

adrian-lison:
> The reason this is failing at the moment is because I believe it has to have a version on the target branch in order to work (so here`develop`.

So you think it will work after we have merged this in develop? If we are scared, we could also create another test branch from this one here, make some changes and test a pull request from the test branch into this one.

test.R
seabbs: Removed in https://github.com/epinowcast/epinowcast/pull/200/commits/f2d6fc03895d1448749f9fec2f7a2ca4c818df35

seabbs:
As suggested by @adrian-lison this is now being tested in #203

seabbs:
Proving to be a bit of a fiddle to get right.

adrian-lison:
@seabbs I looked into this now and it seems that the current point of failure is really the function `pin_assets`. It works neither when just using `"touchstone"` or just using `"inst/examples"`, and the failure happens somewhere in the call to `pin_assets` (see [here](https://github.com/lorenzwalthert/touchstone/blob/e4b325587c6425f00fbf6dd989132bd50532a9e2/R/utils.R)), which then triggers the following error from the `cli` package ([see here](https://rdrr.io/github/r-pkgs/boxes/src/R/ansi-hyperlink.R)):
```
Error in vapply(X, FUN, FUN.VALUE = character(1), ..., USE.NAMES = USE.NAMES) : 
  values must be type 'character',
 but FUN(X[[1]]) result is type 'list'
Calls: <Anonymous> ... transform -> make_link -> make_link_file -> vcapply -> vapply
```

I haven't figured out yet which call exactly it is (maybe more deep down the hierarchy, because all the direct calls to `cli` in the function seem alright) and what is causing this.

One general guess though I had might be versioning of `{touchstone}`. At the moment, we have this in the workflow
`- uses: lorenzwalthert/touchstone/actions/receive@main`
but maybe we have to specify the correct version or use the v1 wildcard: https://github.com/lorenzwalthert/touchstone/blob/main/actions/README.md ?

adrian-lison:
My best guess is that one or several of the directories that we want to pin cannot be found and then it is doing something really weird when trying to write a warning on this using `cli`, see [here](https://github.com/lorenzwalthert/touchstone/blob/e4b325587c6425f00fbf6dd989132bd50532a9e2/R/utils.R#L213)
```
if (!all(valid_dirs)) {
    cli::cli_warn(paste0(
      "The following asset{?s} could not be found and will ",
      "not be copied: {.path {dirs[!valid_dirs]}}"
    ))
  }
  ```
  
  somehow, if trying to format a path which is not a character but a list, then we get exactly the error we have. But I have no idea how this could happen, I cannot reproduce at this function level.

adrian-lison:
@seabbs It must really be a version issue though. Did you know that `{touchstone}` is recently also failing with the same error in `EpiNow2`? https://github.com/epiforecasts/EpiNow2/actions/runs/4366678150/jobs/7637021308?pr=372

So maybe we can go back to an older version of touchstone? And we should potentially inform the developer.

adrian-lison:
Tried a workaround that avoides `pin_assets` by sourcing the relevant files directly (works as long they agree in all branches), but now we get new errors. Currently I am stuck with [this one](https://github.com/epinowcast/epinowcast/actions/runs/4501160645/jobs/7921222512?pr=203). Very weird.

seabbs:
> but maybe we have to specify the correct version or use the v1 wildcard:

Really good point about using standard versions. This seems like it might well be it (especially due to the change in `EpiNow2` from working -> not. 

> somehow, if trying to format a path which is not a character but a list, then we get exactly the error we have. But I have no idea how this could happen, I cannot reproduce at this function level.

Great job digging and yeah I have no idea what is happening here. 

By the way for loca testing I have been finding `act` really helpful for GitHub Actions: https://github.com/nektos/act

seabbs:
> ut now we get new errors. Currently I am stuck with [this one](https://github.com/epinowcast/epinowcast/actions/runs/4501160645/jobs/7921222512?pr=203). Very weird.

This is very weird

seabbs:
> but maybe we have to specify the correct version or use the v1 wildcard:

It looks like the last release of this was in Jan 2022 but a bug was fixed due to later releases of `cli` in November 2022. This could maybe be an issue. I have reverted to use `main` for both actions.

seabbs:
So by comparing with working version @sbfnk implemented for `EpiNow2` I have got this working. The core of the issue appears to have been trying to source within the benchmark. Moving to just a function call appears to have got things working (🤞🏻)

seabbs:
All finally working now so just a matter of fine-tuning what we want to run in the benchmarks. Annoyingly in needs to recompile the model each time and so this slows everything down. It doesn't seem like we can avoid this either.

seabbs (commented):
This all appears to be working and I think the current benchmarks are a good MVP. Planning on merging as a self-review and we can iterate.

Issue #206, opened 2023-03-23
Depreciate enw_dates_to_factors


seabbs:
We currently export a function `enw_dates_as_factors` that converts dates to factors and drops all other variables. As far as I can tell it is unused in our current code base and also seems unlikely to be needed by the user? 

Given this, I think we can make the case for soft depreciating it to reduce technical debt. 
Use case: 

```r
data <- data.frame(date = as.Date("2019-01-01") + 0:2, x = 1:3)
enw_dates_to_factors(data)
```

adrian-lison:
You probably mean `enw_dates_to_factors`, right?

This was before my time, but apparently you only it used for the following:
```
enw_day_model <- function(metaobs, rw = FALSE) {

  # turn dates into factors
  metaobs <- enw_dates_to_factors(metaobs)

  # build effects design matrix (with  no contrasts)
  fixed <- enw_design(~date, metaobs, no_contrasts = TRUE)

 ...
}
```

I can't think of a need for this atm.

PR #208, opened 2023-03-25
Extend documentation and tests for pre and post processing

seabbs:
This PR adds tests and documentation for all pre and post-processing functions for which this was missing. It also adds examples where these were missing for documented functions. Note that the documentation and tests are all a first pass and will likely need improvement over time.

In making these examples it also adds support for no`.group` variable in all preprocessing functions via the addition of a new internal helper function `add_group()`. 

Finally it resolves the spurious test warnings for snapshot tests which were linked to unstated formatting requirements. 

**In going through the pre/post processing I think both could do with a rethink/refactor in the future.**


R/preprocess.R
```diff
@@ -199,63 +244,101 @@ enw_extend_date <- function(metaobs, days = 20, direction = "end") {
...
+    stop(
+      "The `.group` column is reserved for internal use. Please remove it ",
+      "from your data before calling `enw_assign_group`."
+    )
   }
-  obs <- data.table::as.data.table(obs)
-  if (length(by) == 0) {
-    obs <- obs[, .group := 1]
-  } else {
+  if (length(by) != 0) {
```
pearsonca, line 283: hmm - feels a bit like the `by` argument should be in add_group. I'm okay with this function overall as a wrapper to the capability (e.g. to manage argument checking that isn't typically done on internal function), but a bit odd this doesn't live in `add_group`.

R/preprocess.R
```diff
@@ -498,7 +581,8 @@ enw_incidence_to_cumulative <- function(obs, by = c()) {
 #' obs <- enw_example("preprocessed")$obs[[1]]
 #' enw_delay_filter(obs, max_delay = 2)
 enw_delay_filter <- function(obs, max_delay) {
-  obs <- data.table::copy(obs)
+  obs <- data.table::as.data.table(obs)
+  obs <- add_group(obs)
```
pearsonca, line 585: ibid

R/check.R
```diff
@@ -76,6 +76,20 @@ check_group <- function(obs) {
   return(invisible(NULL))
 }
 
+#' Add a reserved grouping variable if missing
+#'
+#' @param x A data frame or data table
```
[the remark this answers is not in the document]
pearsonca: re my later comments - here, add_group doesn't *necessary* create a copy, but it's guaranteed to if it needs to. seems better to have it guarantee copy created (since it seems to always be used alongside a guarantee to create a copy).

R/preprocess.R
```diff
@@ -199,63 +244,101 @@ enw_extend_date <- function(metaobs, days = 20, direction = "end") {
   exts[, observed := FALSE]
 
   exts <- rbind(
-    data.table::copy(metaobs)[, observed := TRUE],
+    metaobs[, observed := TRUE],
     exts[, observed := FALSE]
```
pearsonca, line 248: isn't this repeating step from a few lines above?

tests/testthat/helper-functions.R
```diff
@@ -24,3 +24,9 @@ skip_on_local <- function() {
   }
   skip("Not on CI")
 }
+
+round_numeric <- function(dt) {
+  cols <- colnames(dt)[purrr::map_lgl(dt, is.numeric)]
+  dt <- dt[, (cols) := lapply(.SD, round, 0), .SDcols = cols]
+  return(dt)
+}
```
pearsonca, line 32: might be a bit clearer as `round_numerics`. Also, `sapply` could replace `purrr::map_lgl`, don't need to specify the default 0 (though okay with telegraphing intent).

tests/testthat/test-add_group.R
```diff
@@ -0,0 +1,13 @@
...
+    data.table::data.table(x = 1, .group = 1)
+  )
+})
+
+test_that("add_group doesn't add a group when one is present", {
+  expect_equal(
+    epinowcast:::add_group(data.frame(x = 1, .group = 1)),
+    data.frame(x = 1, .group = 1)
+  )
+})
```
pearsonca, line 13: should test this with a different pre-existing group, since this is the group that would be created anyway.

tests/testthat/test-enw_assign_group.R
pearsonca, line 1: if you incorporated suggestion about including by argument to add_group, some of this testing would also be duplicated for add_group (though should remain here, as this is the user surface area).

should there also be error / warning if "by" argument isn't present in data?

tests/testthat/test-enw_dates_to_factors.R
pearsonca, line 1: based on this, seems like answer to my earlier Q is that dropping columns is intended; should clarify that in interface documentation.

pearsonca (commented):
Overall seems fine, assorted TODOs in detailed notes. Summary:
 - regularize some documentation style that bounces around on type (my recommendation: link to other documentation on the type + fixed with code annotation)
 - opportunity to be a bit smarter on `add_group` implementation + use
 - need to establish a principle on what you offer the users - stick to data.frame-only guarantee (pro: leaves you internal flexibility by limiting commitment, con: data.frames suck) vs make a consistent data.table guarantee (mirror pro/con).

R/check.R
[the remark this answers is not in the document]
seabbs: I think we want the required copy to be specified in top-level functions vs in `add_group`. For the documentation of `data.frame` I think it might be better to only ask for `data.frame` vs anything that inherits from it as anything based on a `data.frame` should be coerce-able using [data.table::as.data.table()] internally (apart from perhaps some edge cases).

R/model-design-tools.R
```diff
@@ -1,39 +1,48 @@
...
 is.Date <- function(x) {
   # nolint
   inherits(x, "Date")
 }
 
-#' @title FUNCTION_TITLE
-#' @description FUNCTION_DESCRIPTION
-#' @param data PARAM_DESCRIPTION
-#' @return OUTPUT_DESCRIPTION
+#' @title Convert all Dates to Factors
```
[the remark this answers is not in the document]
seabbs: Is that supported by roxygen?

R/model-design-tools.R
```diff
@@ -1,39 +1,48 @@
...
+#' @description Converts all Date columns to factors
+#' in a `data.frame`.
+#' @param data A `data.frame`.
```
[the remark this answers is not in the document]
seabbs: Yes we don't want to force data.table on people (though the output all being `data.table` does do that to some extent at the moment. Agree this can be added to the contributor guide but that feels like its own issue.

R/model-design-tools.R
```diff
@@ -50,24 +59,43 @@ mod_matrix <- function(formula, data, sparse = TRUE, ...) {
...
+#' number of rows in the design matrix and an index vector that
+#' allows the full design matrix to be reconstructed. This is useful
+#' for models that have many repeated rows in the design matrix and that
+#' are computationally expensive to fit. This function also allows
+#' for the specification of contrasts for categorical variables.
+#'
+#' @param formula An R formula.
 #'
-#' @param data PARAM_DESCRIPTION
+#' @param data A data frame containing the variables in the formula.
```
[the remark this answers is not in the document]
seabbs: agree - will search through and make consistent

R/model-design-tools.R
```diff
@@ -119,58 +147,114 @@ enw_design <- function(formula, data, no_contrasts = FALSE, sparse = TRUE,
...
+#' @param effects A data.table with the following columns:
+#' - effects: the name of the effect
+#' - fixed: a logical indicating whether the effect is fixed (1) or random (0).
 #'
-#' @param string PARAM_DESCRIPTION
+#' This is the output of [enw_effects_metadata()].
 #'
-#' @param var_name PARAM_DESCRIPTION, Default: 'sd'
+#' @param string A string that will be used to identify the effects that
+#' should be pooled. This can be a regular expression.
```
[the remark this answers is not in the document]
seabbs: its just passed to the lower level function so would depend on what that is.

R/model-design-tools.R
```diff
@@ -119,58 +147,114 @@ enw_design <- function(formula, data, no_contrasts = FALSE, sparse = TRUE,
...
 #'
-#' @param finder_fn PARAM_DESCRIPTION, Default: startsWith
-#' @return OUTPUT_DESCRIPTION
+#' @param var_name The name of the new column that will be added to the
+#' `effects` data.table. This column will be 1 for the effects that match the
+#' string and 0 otherwise. Defaults to 'sd'.
+#'
+#' @param finder_fn A function that will be used to find the effects that
```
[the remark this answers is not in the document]
seabbs: yes and agree. this massive dump is very much a first pass!

R/model-design-tools.R
```diff
@@ -119,58 +147,114 @@ enw_design <- function(formula, data, no_contrasts = FALSE, sparse = TRUE,
...
 #' @export
+#' @examples
+#' metaobs <- data.frame(week = 1:3)
+#' enw_add_cumulative_membership(metaobs, "week")
+#'
+#' metaobs <- data.frame(week = 1:3, .group = c(1,1,2))
+#' enw_add_cumulative_membership(metaobs, "week")
 enw_add_cumulative_membership <- function(metaobs, feature) {
   metaobs <- data.table::as.data.table(metaobs)
+  metaobs <- add_group(metaobs)
```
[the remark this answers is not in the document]
seabbs: yeah exactly, just going for clarity at the moment vs speed as its so tiny vs the fitting part of the package. I think we could do a big pass across the package on this in another PR.

R/postprocess.R
```diff
@@ -130,27 +157,41 @@ enw_nowcast_samples <- function(fit, obs) {
...
 enw_summarise_samples <- function(samples, probs = c(
                                     0.05, 0.2, 0.35, 0.5,
                                     0.65, 0.8, 0.95
                                   ),
-                                  by = c("reference_date", ".group")) {
-  obs <- samples[.draw == 1]
-  obs[, c(".draw", ".iteration", "sample", ".chain") := NULL]
+                                  by = c("reference_date", ".group"),
+                                  link_with_obs = TRUE) {
+  obs <- samples[.draw == min(.draw)]
```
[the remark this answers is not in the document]
seabbs: yeah agree. In general the post-processing leaves a lot to be desired.

R/postprocess.R
```diff
@@ -169,23 +210,38 @@ enw_summarise_samples <- function(samples, probs = c(
...
-  summary <- purrr::reduce(
-    list(obs, summary, quantiles), merge,
-    by = by
-  )
+  if (link_with_obs) {
+    dts <- list(obs, summary, quantiles)
+  } else {
+    dts <- list(summary, quantiles)
+  }
+  summary <- purrr::reduce(dts, merge, by = by)
```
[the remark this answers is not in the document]
seabbs: it does as otherwise this is a breaking change. Style wise we aren't using same line if statements but will commit a DRY version

R/preprocess.R
```diff
@@ -1,24 +1,51 @@
...
+#' @examples
+#' obs <- data.frame(reference_date = as.Date("2021-01-01"), x = 1:10)
+#' enw_metadata(obs, target_date = "reference_date")
 enw_metadata <- function(obs, target_date = c(
                            "reference_date", "report_date"
                          )) {
+  obs <- data.table::as.data.table(obs)
   choices <- c("reference_date", "report_date")
   target_date <- match.arg(target_date)
   date_to_drop <- setdiff(choices, target_date)
```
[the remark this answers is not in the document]
seabbs: Agree - this feels like its own issue vs for this PR.

R/check.R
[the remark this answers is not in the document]
pearsonca: mmm, then have `add_group` require a data.table argument and expressly modify that object?

R/model-design-tools.R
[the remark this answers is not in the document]
pearsonca: code in the title? pretty sure yes?

R/model-design-tools.R
```diff
@@ -1,39 +1,48 @@
...
   # nolint
   inherits(x, "Date")
 }
 
-#' @title FUNCTION_TITLE
-#' @description FUNCTION_DESCRIPTION
-#' @param data PARAM_DESCRIPTION
-#' @return OUTPUT_DESCRIPTION
+#' @title Convert all Dates to Factors
+#' @description Converts all Date columns to factors
```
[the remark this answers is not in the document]
pearsonca: (my understanding from tests later indicates it's intended, so probably modify docs to reflect)

R/model-design-tools.R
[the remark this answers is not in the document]
seabbs: its supported but looks pretty werid in my editor keeping for now.

tests/testthat/test-enw_dates_to_factors.R
seabbs: done. Note that #206 exists (not for this PR). I really don't see the need for this function.

seabbs:
Thanks for the review @pearsonca - some good points. 

> need to establish a principle on what you offer the users - stick to data.frame-only guarantee (pro: leaves you internal flexibility by limiting commitment, con: data.frames suck) vs make a consistent data.table guarantee (mirror pro/con).

I agree. I think we should handle this in another issue/PR though as it goes beyond these changes. My preference is not to force `data.table` on people but to accept that our output can be `data.table` (a bit of a half way house). Docs wise finding a standard way of saying that seems like a good idea. 

> opportunity to be a bit smarter on add_group implementation + use

I agree we don't need to support anything apart from in place `data.table` here so have updated. In terms of smarter use I think sticking to copying happening explicitly at the top level is nice where possible (though I think `check_dates` currently violates this). 

> regularize some documentation style that bounces around on type (my recommendation: link to other documentation on the type + fixed with code annotation)

I think I have mostly done this. Just doing another pass and then will ping for another review.

tests/testthat/test-enw_assign_group.R
seabbs: I think we don't need to make add_group any more complex but agree adding a by arg check is a good idea.

R/model-design-tools.R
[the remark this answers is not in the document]
pearsonca: what do you think about using `...` args here? basically: drop `string` arg, make `...` be passed to `finder_fn`? downside is that your default finder_fn isn't guaranteed to be passed its required args, but that kinda could already happen?

`...` generally should be used with caution, but this seems like the natural case for them?

tests/testthat/test-enw_assign_group.R
seabbs: > should there also be error / warning if "by" argument isn't present in data?

Agree. I have added a new internal `check_by` function.

R/model-design-tools.R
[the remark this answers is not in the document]
seabbs: Resolving here. Should move to an issue.

R/model-design-tools.R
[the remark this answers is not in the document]
seabbs: used find and replace.

R/utils.R
```diff
@@ -4,6 +4,16 @@
 #' @importFrom stats median rnorm
 NULL
 
+#' @title Check a vector is a Date
```
pearsonca, line 7: missed a spot.

R/preprocess.R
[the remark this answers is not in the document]
seabbs: (also I don't know how to do this without some play)

R/preprocess.R
seabbs: as in `enw_assign_group` and `add_group` could be merged? I agree  (though the does it allow a `.group` var to exist is currently different and needs to be as it is currently used). I think we should discuss in another issue vs this PR though (as here trying to focus on docs).

R/preprocess.R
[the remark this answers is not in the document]
pearsonca: Look at `match.arg` at an R prompt - will show you the function body (though agree the reflection code is a ...thing to grok). Agree can be carved as own small TODO (good-first-issue style).

R/model-design-tools.R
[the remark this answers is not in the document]
seabbs: yeah I think this could be a good. Idea. Its a bit of a breaking change though and I think should handle in a new issue/PR.

seabbs:
I think this is ready for another review now. A few of your points seem like they could be better handled in a new issue/PR vs here so have left those unresolved.

R/preprocess.R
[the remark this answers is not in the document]
seabbs: Now discussed in #212

pearsonca (changes requested):
I think the find-replace got muddled.

R/check.R
```diff
@@ -78,7 +78,7 @@ check_group <- function(obs) {
 
 #' Check by variables are present in the data
 #'
-#' @param obs An observation data.frame.
+#' @param obs An observation `data.frame``.
```
pearsonca, line 81:
```suggestion
#' @param obs An observation `data.frame`.
```

R/formula-tools.R
```diff
@@ -558,7 +558,7 @@ construct_re <- function(re, data) {
 #' effects, random effects using [lme4] syntax (see [re()]), and random walks
 #' defined using the [rw()] helper function.
 #'
-#' @param data A `data.frame` of observations. It must include all
+#' @param data A  `data.frame` of observations. It must include all
```
pearsonca, line 561:
```suggestion
#' @param data A `data.frame` of observations. It must include all
```

R/methods.R
```diff
@@ -50,13 +50,14 @@ summary.epinowcast <- function(object, type = c(
...
 
+
 #' Plot method for epinowcast
 #'
 #' @description `plot` method for class "epinowcast".
 #'
 #' @param x A `data.table` of output as produced by [epinowcast()].
 #'
-#' @param latest_obs A `data.frame` of observed data which may be passed to
+#' @param latest_obs A  `data.frame` of observed data which may be passed to
```
pearsonca, line 60:
```suggestion
#' @param latest_obs A `data.frame` of observed data which may be passed to
```

R/model-design-tools.R
```diff
@@ -1,8 +1,8 @@
 #' @title Convert all `Date`s to `Factor`s
 #' @description Converts all `Date` columns to `Factor`s.
-#' in a `data.frame`.
-#' @param data A `data.frame`.
-#' @return A `data.frame` with all `Date` variables converted to `Factor`s and
+#' in a  `data.frame`.
+#' @param data A  `data.frame`.
+#' @return A  `data.frame` with all `Date` variables converted to `Factor`s and
```
pearsonca, line 5:
```suggestion
#' in a `data.frame`.
#' @param data A `data.frame`.
#' @return A `data.frame` with all `Date` variables converted to `factor`s and
```

R/model-module-helpers.R
```diff
@@ -1,10 +1,10 @@
 #' Identify report dates with complete (i.e up to the maximum delay) reference
 #' dates
 #'
-#' @param new_confirm `new_confirm``data.frame`output from
+#' @param new_confirm `new_confirm` `data.frame` output from
```
pearsonca, line 4: not sure if this is wrong or not. seems fishy.

R/model-module-helpers.R
```diff
@@ -21,23 +21,23 @@ enw_reps_with_complete_refs <- function(new_confirm, max_delay, by = c()) {
 
 #' Construct a lookup of references dates by report
 #'
-#' @param missing_reference `missing_reference``data.frame`output from
+#' @param missing_reference `missing_reference` `data.frame` output from
```
pearsonca, line 24: also fishy.

R/model-module-helpers.R
```diff
@@ -21,23 +21,23 @@ enw_reps_with_complete_refs <- function(new_confirm, max_delay, by = c()) {
...
 #' [enw_preprocess_data()].
 #'
-#' @param reps_with_complete_refs A `data.frame` of report dates with complete
+#' @param reps_with_complete_refs A  `data.frame` of report dates with complete
 #' (i.e fully reported) reference dates as produced using
 #' [enw_reps_with_complete_refs()].
 #'
-#' @param metareference `metareference``data.frame`output from
+#' @param metareference `metareference` `data.frame` output from
```
pearsonca, line 31: also fishy

R/model-module-helpers.R
```diff
@@ -21,23 +21,23 @@ enw_reps_with_complete_refs <- function(new_confirm, max_delay, by = c()) {
 
 #' Construct a lookup of references dates by report
 #'
-#' @param missing_reference `missing_reference``data.frame`output from
+#' @param missing_reference `missing_reference` `data.frame` output from
 #' [enw_preprocess_data()].
 #'
-#' @param reps_with_complete_refs A `data.frame` of report dates with complete
+#' @param reps_with_complete_refs A  `data.frame` of report dates with complete
```
pearsonca, line 27:
```suggestion
#' @param reps_with_complete_refs A `data.frame` of report dates with complete
```

R/model-module-helpers.R
```diff
@@ -71,7 +71,7 @@ enw_reference_by_report <- function(missing_reference, reps_with_complete_refs,
 }
 #' Convert latest observed data to a matrix
 #'
-#' @param latest `latest``data.frame`output from [enw_preprocess_data()].
+#' @param latest `latest` `data.frame` output from [enw_preprocess_data()].
```
pearsonca, line 74: fishy

R/model-modules.R
```diff
@@ -14,7 +14,7 @@
 #'
 #' @param non_parametric A formula (as implemented in [enw_formula()])
 #' describing the non-parametric logit hazard model. This can use features
-#' defined by reference date and by delay. It draws on a linked `data.frame`
+#' defined by reference date and by delay. It draws on a linked  `data.frame`
```
pearsonca, line 17:
```suggestion
#' defined by reference date and by delay. It draws on a linked `data.frame`
```

R/model-tools.R
```diff
@@ -70,9 +70,9 @@ enw_formula_as_data_list <- function(formula, prefix,
   return(data)
 }
 
-#' Convert prior`data.frame`to list
+#' Convert prior `data.frame` to list
 #'
-#' Converts priors defined in a `data.frame` into a list
+#' Converts priors defined in a  `data.frame` into a list
```
pearsonca, line 75:
```suggestion
#' Converts priors defined in a `data.frame` into a list
```

R/model-tools.R
```diff
@@ -229,11 +229,11 @@ write_stan_files_no_profile <- function(stan_file, include_paths = NULL,
 #' supplied by the user.
 #'
 #' @param diagnostics Logical, defaults to `TRUE`. Should fitting diagnostics
-#' be returned as a `data.frame`.
+#' be returned as a  `data.frame`.
```
pearsonca, line 232:
```suggestion
#' be returned as a `data.frame`.
```

R/model-tools.R
```diff
@@ -229,11 +229,11 @@ write_stan_files_no_profile <- function(stan_file, include_paths = NULL,
...
 #'
 #' @param ... Additional parameters passed to the `sample` method of `cmdstanr`.
 #'
-#' @return A `data.frame` containing the `cmdstanr` fit, the input data, the
+#' @return A  `data.frame` containing the `cmdstanr` fit, the input data, the
```
pearsonca, line 236:
```suggestion
#' @return A `data.frame` containing the `cmdstanr` fit, the input data, the
```

R/plot.R
```diff
@@ -16,10 +16,10 @@ enw_plot_theme <- function(plot) {
 
 #' Generic quantile plot
 #'
-#' @param obs A `data.frame` of summarised posterior estimates
+#' @param obs A  `data.frame` of summarised posterior estimates
```
pearsonca, line 19:
```suggestion
#' @param obs A `data.frame` of summarised posterior estimates
```

R/plot.R
```diff
@@ -83,7 +83,7 @@ enw_plot_obs_by_report <- function(obs, log = TRUE, ...) {
 
 #' Generic quantile plot
 #'
-#' @param posterior A `data.frame` of summarised posterior estimates
+#' @param posterior A  `data.frame` of summarised posterior estimates
```
pearsonca, line 86:
```suggestion
#' @param posterior A `data.frame` of summarised posterior estimates
```

R/plot.R
```diff
@@ -114,7 +114,7 @@ enw_plot_quantiles <- function(posterior, latest_obs = NULL, log = FALSE, ...) {
 
 #' Plot nowcast quantiles
 #'
-#' @param nowcast A `data.frame` of summarised posterior nowcast
+#' @param nowcast A  `data.frame` of summarised posterior nowcast
```
pearsonca, line 117:
```suggestion
#' @param nowcast A `data.frame` of summarised posterior nowcast
```

R/plot.R
```diff
@@ -142,7 +142,7 @@ enw_plot_nowcast_quantiles <- function(nowcast, latest_obs = NULL,
 
 #' Plot posterior prediction quantiles
 #'
-#' @param pp A `data.frame` of summarised posterior predictions
+#' @param pp A  `data.frame` of summarised posterior predictions
```
pearsonca, line 145:
```suggestion
#' @param pp A `data.frame` of summarised posterior predictions
```

R/preprocess.R
```diff
@@ -5,7 +5,7 @@
 #' (reference or report), `confirm`, `max_confirm``, and `cum_prop_reported`
 #' are dropped and the first observation for each group and date is retained.
 #'
-#' @param obs A`data.frame`or data.table with columns: `reference_date` and/or
+#' @param obs A `data.frame` or data.table with columns: `reference_date` and/or
```
pearsonca, line 8:
```suggestion
#' @param obs A `data.frame` or `data.table` with columns: `reference_date` and/or
```

R/preprocess.R
```diff
@@ -195,7 +195,7 @@ enw_add_metaobs_features <- function(metaobs,
 #' for nowcasting purposes or to include additional dates for backcasting
 #' when using a renewal process as the expectation model.
 #'
-#' @param metaobs A `data.frame` with a `date` column.
+#' @param metaobs A  `data.frame` with a `date` column.
```
pearsonca, line 198:
```suggestion
#' @param metaobs A `data.frame` with a `date` column.
```

R/preprocess.R
```diff
@@ -255,7 +255,7 @@ enw_extend_date <- function(metaobs, days = 20, direction = "end") {
 #' be assigned a unique group. If `by` is not specified, then all rows
 #' will be assigned to the same group.
 #'
-#' @param obs A data.table or`data.frame`without a `.group` column.
+#' @param obs A data.table or `data.frame` without a `.group` column.
```
pearsonca, line 258:
```suggestion
#' @param obs A `data.table` or `data.frame` without a `.group` column.
```

R/preprocess.R
pearsonca: sort of, but not quite - agree another issue. basically, focus the pure technical capabilities in add_group (such as executing on a by clause), but have it assume (because it's internal function) that it's not handed garbage. the enw_assign_group handles all the crap users might throw at it.

pearsonca: but this assumes you want to leverage those technical capabilities in other places you're using add_group, not just enw_assign_group

pearsonca (commented):
Soooooo...probably want to squash when doing this merge.

R/model-module-helpers.R
seabbs: its correct but perhaps not clear `new_confirm` is a `data.frame` output as part of the `epinowcast` output.

R/model-module-helpers.R
seabbs: same as above

R/model-module-helpers.R
seabbs: same as above

R/model-module-helpers.R
seabbs: same as above.

seabbs:
> Soooooo...probably want to squash when doing this merge.


I did a great job on that manual review. Nice work with the suggested changes. In general, I am actually a really big fan of using squash merges for PRs.

R/preprocess.R
seabbs: ah got you. I think we always want to go via assign group (at least for now) but agree we could have assign group be the exposed to the user version and add_group be the less safe internal use version (though for people inspecting the code this may lead them astray (if for example they want to role their own version of the preprocessing function for some reason).

R/model-design-tools.R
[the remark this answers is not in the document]
seabbs: I've added a new issue for this (#213).

R/preprocess.R
seabbs: Could you draft up a new issue for this and will discuss/prototype ahead of the next release? I think I need to think about it a little.

pearsonca (approved):

R/preprocess.R
seabbs: Added a new issue (#214) to discuss this.

seabbs:
Thanks for the detailed and helpful review @pearsonca ❤️

PR #209, opened 2023-03-26
Add examples and tests for `epinowcast` methods and uncovered plots

seabbs:
This PR adds basic examples and tests for `summary.epinowcast()`, `plot.epinowcast()`, and `enw_plot_obs()` (which was the only plotting function not fully covered by the new tests for `plot.epinowcast()`). 

It also resolves an issue with the default in `enw_plot_quantiles` which was `TRUE` vs `FALSE` as documented.


R/methods.R
```diff
@@ -61,6 +75,16 @@ summary.epinowcast <- function(object, type = c(
...
+#' nowcast <- enw_example("nowcast")
+#' latest_obs <- enw_example("obs")
+#'
+#' # Plot nowcast
+#' plot(nowcast, latest_obs = latest_obs, type = "nowcast")
+#'
+#' # Plot posterior predictions by reference date
+#' plot(nowcast, type = "posterior_prediction") +
+#'  ggplot2::facet_wrap(ggplot2::vars(reference_date), scales = "free")
 plot.epinowcast <- function(x, latest_obs = NULL, type = c(
```
pearsonca, line 88: arguably, `ggplot.epinowcast` is more appropriate here.

pearsonca: also, to emphasize previously offered opinion: the plotting stuff should be a separate package.

R/plot.R
```diff
@@ -77,14 +86,20 @@ enw_plot_obs_by_report <- function(obs, log = TRUE, ...) {
...
+#' @seealso [enw_plot_nowcast_quantiles()], [enw_plot_pp_quantiles()]
 #' @family plot
 #' @inheritParams enw_plot_obs
 #' @export
-enw_plot_quantiles <- function(posterior, latest_obs = NULL, log = TRUE, ...) {
+#' @examples
+#' nowcast <- enw_example("nowcast")
+#' nowcast <- summary(nowcast, probs = c(0.05, 0.2, 0.8, 0.95))
+#' enw_plot_quantiles(nowcast, x = reference_date)
+enw_plot_quantiles <- function(posterior, latest_obs = NULL, log = FALSE, ...) {
```
pearsonca, line 102: feels weird. i get this is MVP, but: quantiles should be discovered, rather than hard coded.

pearsonca (commented):
Seems generally fine, but also: this should be its own extension package.

R/methods.R
seabbs: Yes I agree. This depends on developer time and I think is only a good idea when moving out of the experimental stage. If you go multi-repo to early you can tie yourself into a dev nightmare. 

The plan is to do this with all pre and post processing at some point (just not yet)

seabbs: > arguably, ggplot.epinowcast is more appropriate here.

Agree, but its very standard practice to use `ggplot2` inside `plot` calls.

R/plot.R
seabbs: That is what I did in `EpiNow2` and it was a very big pain to dev + maintain. I am happy to see this here though I just don't want to do the dev (at least right now).

seabbs: (happy to see an issue on this of course)

seabbs:
@pearsonca do you have a review decision on this? Going to merge in #208 first, resolve merge issues, and then merge this in once ready.

pearsonca (approved):

pearsonca:
Comfortable with it. Added issue(+) for splitting off vis package + introspecting Qs.

Issue #211, opened 2023-03-26
Required object types for users


pearsonca:
There's a general engineering principle to minimize constraints on external interfaces, as well as minimizing *promised* capabilities (at least until there's an identified demand for a capability).

Should pursue that for epinowcast generally. Known TODOs:
 - [ ] `data.frame` vs `data.table` inputs + documentation
 - [ ] `data.table` outputs + documentation
 - [x] CONTRIBUTING guidelines to reflect this principle
 - [ ] review for other interface items potentially violating this principle

In general, we should only require `data.frame` inputs (n.b. this of course includes `data.table` inputs). In many cases, we could allow *more* inputs (because `as.data.table` will handle them), but that would lock us into that more complicated contract.

However, because the typical yields are `data.table`s and they behave slightly (but importantly) differently from `data.frame`s (and we don't want to recast everything to `data.frame`), we need to tell that to the users.

pearsonca:
i think it's fine to return `data.table`: it "is a" `data.frame` with the narrow exception of pass-by-reference to functions, so people can toss into their "tidy" pipelines with aplomb.

the key is just to consistently tell people that's what they're getting, and probably have somewhere prominent a warning label / reminder (the README, package docs, vignette - but probably not on every single return doc).

seabbs:
I'm not sure I have the mental fortitude to deal with this issue in this release... Do we think its critical we get this standardised now or can we push?

pearsonca:
I think just the style guide => then gradual enforcement.

seabbs:
I have split out tasks into their own issues in more addressable formats see #263 and #262. I have also addressed the syle guide in #264 which will close this issue when merged (in favour of #263 and #262).

PR #222, opened 2023-04-04
draft change: add_pooling_effect to more generic finder_fn

pearsonca:
still need some other `finder_fn` examples to add to testing?

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: epinowcast
 Title: Flexible Hierarchical Nowcasting
-Version: 0.2.0.6000
+Version: 0.2.0.7000
 Authors@R:
   c(person(given = "Sam Abbott",
            role = c("aut", "cre"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,9 @@
+# epinowcast 0.2.1
+
+## Package
+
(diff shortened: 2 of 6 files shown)


seabbs (changes requested):
This all looks good to me and as suggested is certainly clearer. Aside from another test case, which we can sort of dream up I think, it is important we have a really clear news note here indicating that this is a breaking change. I actually wonder if we should have a news section for breaking changes specifically just to highlight this.

pearsonca:
added another test to highlight `...` capability + drafted NEWS update.

also noted during testing that `enw_add_pooling_effect` side effects it's argument. that's not *generally* the approach used elsewhere in the library, but the documentation does (sorta) frame it that way.

pearsonca:
ugh, so this kinda goes beyond the limited scope of the issue, but:
should we address applying the interface ethos (this is public, therefore: should be documented as taking a data.frame, should not side effect argument, etc)?
OR should that be addressed as some comprehensive review targetting only that question => changes (+ writing developer guidance for that)?

seabbs:
I just tweaked the proposed news item and incremented the development version. Just doing a final review and then will merge.

> ugh, so this kinda goes beyond the limited scope of the issue, but:
should we address applying the interface ethos (this is public, therefore: should be documented as taking a data.frame, should not side effect argument, etc)?
OR should that be addressed as some comprehensive review targetting only that question => changes (+ writing developer guidance for that)? (i.e. something addressing https://github.com/epinowcast/epinowcast/issues/211)

I think this goes beyond the scope of this issue and we should deal with it in #211 as you suggest.

seabbs (approved):
LGTM

PR #226, opened 2023-04-11
Disable lint workflow on pushes

Bisaloo:
Since this workflow is set up to run on files modified in a PR, it doesn't make sense to have it run in pushes (and it will fail). It's not an issue since all changes are done via PR so you are sure that this workflow will still always run on new code.


seabbs (approved):
Good catch!

PR #229, opened 2023-04-12
Do not use else after stop()

Bisaloo:
In the words of Jenny Bryan: "there is no else, there is only if."

https://speakerdeck.com/jennybc/code-smells-and-feels?slide=45

Having `else` after `return()` of `stop()` increases the number of branches in the code, which makes it harder to read. It also translates into a higher cyclomatic complexity.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: epinowcast
 Title: Flexible Hierarchical Nowcasting
-Version: 0.2.0.9000
+Version: 0.2.0.10000
 Authors@R:
   c(person(given = "Sam Abbott",
            role = c("aut", "cre"),
--- a/NEWS.md
+++ b/NEWS.md
@@ -10,8 +10,8 @@ This is release is in development. It is not yet ready for production use. If yo
 ## Package
 
 - Fixed some typos in `README.md`, `NEWS.md`, the `model.Rmd` vignette and `convolution_matrix()` documentation. The `WORDLIST` used by spelling has also been updated by eliminate false positives. See #221 by @Bisaloo and reviewed by @seabbs and @adrian-lison.
-- Added more non-default linters in `.lintr` configuration file. This file is used when `lintr::lint_package()` is run or in the new `lint-changed-files.yaml` GitHub Actions workflow. See #220 by @Bisaloo and reviewed by @pearsonca and @seeabs.
-- Switched to the `lint-changed-files.yaml` GitHub Actions workflow instead of the regular `lint.yaml` to avoid annotations unrelated to the changes made in the PR. See #220 by @Bisaloo and reviewed by @pearsonca and @seeabs.
+- Added more non-default linters in `.lintr` configuration file. This file is used when `lintr::lint_package()` is run or in the new `lint-changed-files.yaml` GitHub Actions workflow. See #220 by @Bisaloo and reviewed by @pearsonca and @seabbs.
+- Switched to the `lint-changed-files.yaml` GitHub Actions workflow instead of the regular `lint.yaml` to avoid annotations unrelated to the changes made in the PR. See #220 by @Bisaloo and reviewed by @pearsonca and @seabbs.
 - Added tests for `summary.epinowcast()` and `plot.epinowcast()` methods. See #209 by @seabbs and reviewed by @pearsonca.
 - Added tests for `enw_plot_obs()` where not otherwise covered by `plot.epinowcast()` tests. See #209 by @seabbs and reviewed by @pearsonca.
 - Made the `.group` variable optional for all preprocessing functions using a new `add_group()` internal function. See #208 by @seabbs and reviewed by @pearsonca.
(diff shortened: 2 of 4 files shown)


pearsonca:
disagree with this notion as a rule of thumb (and apparently, the cyclomatic complexity metric, if it triggers on this sort of thing).

in this *particular* case, i think the `stopifnot` named argument idiom would be better anyway:

```
stopifnot("error1text"=checkcondition, "error2text"=checkcondition, ...)
```

Bisaloo:
I double checked and indeed, it doesn't affect cyclomatic complexity but I still believe this is a good change that makes the code easier to read.

``` r
cyclocomp::cyclocomp(function() {
  if (a) {
    stop("error") 
  } else {
    message("good")
  }
})
#> [1] 2

cyclocomp::cyclocomp(function() {
  if (a) {
    stop("error") 
  }
  message("good")
})
#> [1] 2
```

<sup>Created on 2023-04-12 with [reprex v2.0.2.9000](https://reprex.tidyverse.org)</sup>

pearsonca:
For error-checking / messages, i think this generally reads more clearly (comments for emphasis - they're unnecessary imo, especially if this idiom is repeated through a library)
```
stopifnot( # ... we're doing some error checking
  "checking this feature" = ..., # against some logical expression
  ... # repeat for all the desired checks
)
```

pearsonca:
Aside: I am on-board with the don't-if-else-your-errors mentality, just not necessarily the `return()` bit. There are two useful idioms for returns in branching code:

 - pre declare the object to manipulate, make the manipulations in the different branches, return after the branches.
 - have each branch return at its end

I have a slight preference for the first, but its not always practical, nor is it always more readable.

pearsonca:
looking at the linked JB example, agree the LHS is code smell, but again: the `stopifnot(...)` idiom would be a much clearer solution than the RHS. That idiom telegraphs: "Here are all the things we can check up front and immediately exit, for reasons".

However, *sometimes* there's actually some branching logic (e.g. when embracing some polymorphism in function arguments, though yes that too can go overboard), and you can't properly check exit conditions until you've made some progress on evaluating arguments.

Bisaloo:
I agree that `stopifnot()` is easier to understand for simple checking. I can do the change for the case at hand.

seabbs:
Everyone agrees it appears the original is not ideal. 

Choice is between using individual if and stop checks or doing a single `stopifnot`. I don't have a strong preference here. @pearsonca could you draft a suggested version using `stopifnot` so we can compare?

pearsonca:
Have y'all stood up style guidelines or some such? I think the practical solution is roughly:

 - do this PR now (agree, generally better + doesn't make move to stopifnot harder)
 - have those style guidelines in place
 - have a issue(s) => PR(s) exclusively focused on implementing guidance item(s), including this one

seabbs:
> Have y'all stood up style guidelines or some such

There are (in the contributing guide) but they are quite vague.

Bisaloo:
Maintaining a thorough style guide is so much work that I usually try to stick with existing ones, even if I disagree with some of their choices (as long as we still agree on most points obviously).

pearsonca:
> Maintaining a thorough style guide is so much work that I usually try to stick with existing ones, even if I disagree with some of their choices (as long as we still agree on most points obviously).

Doesn't have to be complete (e.g. "the specified lintr config passes" covers a *lot*), and this is more of a "style" guide anyway. I *think* it's roughly at the moment:

 - generally, name exported functions with `enw_...`, and *don't* do that for internal functions
 - do check arguments on externally facing functions, don't on internal ones
 - require `data.frame` only for inputs, indicate `data.table` for outputs
 - (as of this discussion) check as many arguments at the outset of `enw_` with `stopifnot(...)` block

Others?

I do think the contributing guidelines currently are too focused on "process" - worthwhile to establish what is the "vocabulary" or "idioms" for enw. That's has some elements of "style" guidelines, but is generally a bit more meta (i.e. human-enforceable only) than concrete (i.e. static analysis works fine).

seabbs:
> Doesn't have to be complete (e.g. "the specified lintr config passes" covers a lot), and this is more of a "style" guide anyway. I think it's roughly at the moment

These are some good points @pearsonca. Shall we move this discussion to an issue ahead of updating the contributing guide? 

@TimTaylor !!! (all good points)

pearsonca:
In terms of actual code, for `check_date`, I'd prefer something akin to:

```r
with(obs, stopifnot(
  "Both `reference_date` and `report_date` must be columns." =
     !(is.null(reference_date) && is.null(report_date)),
  "`reference_date` must be a column." = !is.null(reference_date),
  "`report_date` must be a column." = !is.null(report_date),
 local = ... # not quite sure here; seems like this is an internal function used for checks - parent.frame?
))
```

probably before the internal copy steps as well?

seabbs (approved):
LGTM to me. I have added a news item and bumped the dev version. Porting the outstanding conversation here to two new issues.

seabbs:
#233 captures the discussion here on approaches for improving input checking.

PR #236, opened 2023-04-17
Updates citation file with new authors

seabbs:
This PR closes #230 by updating the `CITATION` file with the new package authors and rerendering the `README`.

PR #237, opened 2023-04-17
Update CITATION to use bibentry

seabbs:
As suggested by @Bisaloo in #236 this PR updates the CITATION file to make use of `bibentry()` as is currently recommended practice.


seabbs (commented):
👯‍♂️

PR #238, opened 2023-04-17
Testing touchstone.

seabbs:
This is a test PR for #200 - DO NOT MERGE.


seabbs:
All finally working now so just a matter of fine-tuning what we want to run in the benchmarks. Annoyingly in needs to recompile the model each time and so this slows everything down. It doesn't seem like we can avoid this either.

Issue #245, opened 2023-04-21
`enw_formula_as_data_list()`s implementation could be improved to better align with DRY principles


pearsonca:
`enw_formula_as_data_list` is a bit convoluted and breaks the DRY (don't repeat yourself) principle:

https://github.com/epinowcast/epinowcast/blob/c1d9bee83bf9c60d5834a485eabf886dfc2087ba/R/model-tools.R#L32-L71

Seems would be clearer as roughly or similar:

```r
# ...initial checking stuff ...
data <- list()
data[c("fintercept", "the other pertinent variables")] <- 0
data[c("fdesign", "these other pertinent variables") <- list(numeric())
if (!missing(formula)) {
  data$fdesign <- formula$fixed$design
   # ... etc
}
names(data) <- sprintf("%s_%s", prefix, names(data))
return(data)
```

Updating this would make the code easier to maintain and expand in the future.

seabbs:
Thanks for this @pearsonca

Small point but when you make this kind of issue can you link to the original code and ideally provide a summary of what the current issue is vs the suggested change (above is fine for me for example but hard to access for many contributors).

I agree the current implementation can be shorten and this looks like a workable way of doing it.

seabbs:
Note that @Lnrivas has a draft PR seeking to address this issue here #249. Potentially help still needed supporting this if interested.

PR #249, opened 2023-04-24
Issue 245: Cleaning enw_formula_as_data_list() (#245)

Lnrivas:
Cleaned enw_formula_as_data_list() as requested in task #245 , please check that it works to your liking so it can be merged asap :)

### DIFF
--- a/R/model-tools.R
+++ b/R/model-tools.R
@@ -31,44 +31,27 @@
 #' enw_formula_as_data_list(prefix = "missing")
 enw_formula_as_data_list <- function(formula, prefix,
                                      drop_intercept = FALSE) {
-  paste_lab <- function(string, lab = prefix) {
-    paste0(lab, "_", string)
-  }
+  data <- list()
+  data[c("fintercept", "fdesign", "fnrow", "findex", "fnindex", "fncol", "rdesign", "rncol")] <- 0
+  data[c("fdesign", "fnrow", "findex", "fnindex", "fncol", "rdesign", "rncol")] <- list(numeric())
   if (!missing(formula)) {
     if (!inherits(formula, "enw_formula")) {
-      stop(
-        "formula must be an object of class enw_formula as produced using
-        enw_formula"
-      )
+      stop("formula must be an object of class enw_formula as produced using enw_formula")
     }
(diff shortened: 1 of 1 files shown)


pearsonca:
To get rid of the linting warnings: try running `lintr::lint_package()` locally, then dealing with the issues (at least those in the code you wrote), re-running, etc until nothing arises from your code.

pearsonca:
The PR process is also to do changes against the `develop` branch, not `main`; see: https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md#development

R/model-tools.R
```diff
@@ -31,45 +31,28 @@
 #' enw_formula_as_data_list(prefix = "missing")
 enw_formula_as_data_list <- function(formula, prefix,
                                      drop_intercept = FALSE) {
-  paste_lab <- function(string, lab = prefix) {
-    paste0(lab, "_", string)
-  }
+  data <- list()
+  data[c("fintercept", "fdesign", "fnrow", "findex", "fnindex", "fncol", "rdesign", "rncol")] <- 0
+  data[c("fdesign", "fnrow", "findex", "fnindex", "fncol", "rdesign", "rncol")] <- list(numeric())
```
pearsonca, line 36: this seems to be the incorrect collection of columns to assign to `numeric()` - copy-paste error?

pearsonca (changes requested):
This is a useful start, but as per comments:

 - please fix the linting issues
 - there also appears to be an assignment error. Did this pass tests? If so, seems like we also need to update the test conditions (to deal with the do-nothing-path).

Thanks for the work so far; looking forward to the update.

R/model-tools.R
pearsonca: Also the wrong columns to 0 - this is on me, for not being clearer in the issue framing.

seabbs:
> The right amount of to-the-point or too clever / terse?

I think this might be the right amount? Thoughts @Lnrivas?

Lnrivas:
Thanks for all the feedback. I will close this pr and make a new one as instructed to correctly work with develop branch and make all changes there.

Issue #234, opened 2023-04-17
Improving the package contributing guide


seabbs:
Our current contributing guide (`CONTRIBUTING.md`) offers some guidance but may be hard to new users to pass and may not give sufficient guidance as to the particular style we have adopted for various areas of our code base. It may also be hard from some users to find in the first place.

In terms of code style I would personally like to stick as closely to default `tidyverse` styles as possible as this makes it easy for new users and allows us to avoid having to maintain extensive custom documentation. 

There are a few areas where we may need more guidance especially when it comes our use of inputing checking (see #233), and `data.table`.

Some comments from contributors (see [here](https://github.com/epinowcast/epinowcast/pull/229) for the original):

## From @pearsonca 

Have y'all stood up style guidelines or some such? I think the practical solution is roughly:

- do this PR now (agree, generally better + doesn't make move to stopifnot harder)
- have those style guidelines in place
- have a issue(s) => PR(s) exclusively focused on implementing guidance item(s), including this one

## From @Bisaloo 

Maintaining a thorough style guide is so much work that I usually try to stick with existing ones, even if I disagree with some of their choices (as long as we still agree on most points obviously).

## From @pearsonca 

> Maintaining a thorough style guide is so much work that I usually try to stick with existing ones, even if I disagree with some of their choices (as long as we still agree on most points obviously).

Doesn't have to be complete (e.g. "the specified lintr config passes" covers a lot), and this is more of a "style" guide anyway. I think it's roughly at the moment:

- generally, name exported functions with enw_..., and don't do that for internal functions
- do check arguments on externally facing functions, don't on internal ones
- require data.frame only for inputs, indicate data.table for outputs
- (as of this discussion) check as many arguments at the outset of enw_ with stopifnot(...) block
Others?

I do think the contributing guidelines currently are too focused on "process" - that's useful, but it's distinctly worthwhile to establish what are the "vocabulary" or "idioms" for enw. That's has some elements of "style" guidelines, but is generally a bit more meta (i.e. human-enforceable only) than concrete (i.e. static analysis works fine).

pearsonca:
Item that came up during `coerce_dt` PR:
 - use of the `..._add_...` in method names suggests side effects. Possible to explicitly embrace that?

seabbs:
This appears to be mostly covered by #264 and so could be closed by it.

PR #264, opened 2023-04-26
Issue 211: Style guide first draft

### LINKED ISSUE #211: Required object types for users

seabbs:
## Description

This PR closes #211.

I have introduced a first pass at a `STYLE_GUIDE.md` which details out input and output requirements as well as some internal code expectations. It clearly states we depend on the `tidyverse` style guide. This new style guide is now linked from the `CONTRIBUTING.md`. 

This PR also partially addresses #234 but more discussion and work may be needed (not for this PR).

## Checklist


CONTRIBUTING.md
```diff
@@ -41,7 +44,7 @@ Please also follow these additional guidelines.
...
-* Format your code according to the [tidyverse style guide](https://style.tidyverse.org/). That formatting can achieved by running `style_pkg()` from [`styler`](https://github.com/r-lib/styler) (which rewrites the files) and `lint_package()` from [`lintr`](https://github.com/jimhester/lintr) (which provides a list of complaints for you to resolve). Note: `styler::style_pkg()` does not examine roxygen content i.e. `@examples` so you will need to check that manually.
+* Format your code according to our [package style](https://github.com/epinowcast/epinowcast/blob/main/STYLE_GUIDE.md) or where not specified the [`tidyverse style` guide](https://style.tidyverse.org/). Automatic `tidyverse` styling can be achieved by running `style_pkg()` from [`styler`](https://github.com/r-lib/styler) (which rewrites the files) and `lint_package()` from [`lintr`](https://github.com/jimhester/lintr) (which provides a list of complaints for you to resolve). Note: `styler::style_pkg()` does not examine roxygen content i.e. `@examples` so you will need to check that manually.
```
pearsonca, line 47: I think this link can be relative; for portability, perhaps do that?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
+# Style Guide
+
+This document outlines the style guide for the `epinowcast` package. This guide is a work in progress and will be updated as the package evolves. We welcome contributions to this guide and encourage you to raise issues or submit PRs if you have any suggestions.
+
+In addition to this guide we also follow the [tidyverse style guide](https://style.tidyverse.org/). This guide is a subset of the `tidyverse` style guide and outlines the additional style requirements for the `epinowcast` package.
+
+## Naming conventions
+
+- We use a `enw_` prefix to delineate functions that are part of the `epinowcast` package. This is to avoid conflicts with other packages and to make it clear to users which functions are part of the package.
```
pearsonca, line 9:
```suggestion
- We use a `enw_` prefix to delineate functions that are exported by the `epinowcast` package. This is to avoid conflicts with other packages and to make it clear to users which functions are part of the package.
```

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+- The use of this prefix is not required for internal functions (i.e. functions that are not exported) or that are unlikely to have naming conflicts with other packages.
+
+## Dependencies
+
+In general we aim to minimise dependencies on other packages where possible. This makes it easier to maintain the package and reduces the risk of breaking changes in other packages impacting our users. However, additional dependencies are sometimes necessary to improve the functionality of the package.
+
+The following guidelines should be followed when using adding dependencies:
+
+- Added to the `Imports` or `Suggests` field `DESCRIPTION` file in alphabetical order.
```
pearsonca, line 18: provide guidance about when imports vs suggests? my rough take:
 - imports: must be present to use the package
 - suggests: may be required to building / test / etc package; also, may provide some extended capability

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+- In the PR that adds the dependency this should be clearly stated in the PR description along with a justification for the dependency, the number and type of downstream dependencies, and an assessment of the risk of the dependency breaking.
+
+More generally when adding functions from external packages (i.e. even if they are already a dependency) the following should be followed:
+
+- Documented in function documentation using the `@importFrom` tag.
+- Used within functions using the `package::function` format.
```
pearsonca, line 24: probably need a bit of a `data.table` caveat here, as we don't use the prefix for some of the functions. Maybe list them?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Input types and checking
+
+- We support inputs that are coercible to `data.table` objects using `data.table::as.data.table()`. This includes `data.frame` and `tibble` objects. This should be clearly documentede in the function documentation.
```
pearsonca, line 28: Not sure about that last sentence overall. I don't think we need to repeat the coerce-able-to guidance - I'd prefer that we simply label inputs as data.frame, outputs as data.table, and have a paragraph somewhere prominent reminding people that both tibbles and data.tables are data.frames, so are suitable inputs.
```suggestion
- We support inputs that are coercible to `data.table` objects using `data.table::as.data.table()`. This includes `data.frame` and `tibble` objects. This should be clearly documented in the function documentation.
```

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+- Any required inputs should be clearly documented in the function documentation.
+- We use an internal function `coerce_dt()` to check inputs are coercible to `data.table` objects and have the correct columns. This function is used in all functions that take data as input. The following function demonstrates this pattern:
+
+```r
+print_dt <- function(dt) {
+    dt <- epinowcast:::coerce_dt(dt, required_cols = c("date", "cases"))
+    return(dt)
```
pearsonca, line 35: probably should mention this pattern as well.
```suggestion
    return(dt[])
```

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+
+- For external facing functions `coerce_dt()` should generally not update by reference (i.e. `copy = TRUE` should be set, the default). In cases where users may benefit from updating by reference the external function should pass through the `copy` argument to `coerce_dt()`.
+- For internal functions `coerce_dt()` should generally update by reference (i.e. `copy = FALSE` should be set) when used internally. This is to avoid unnecessary copying of data.
+
+## Internal data manipulation
+
+- `data.table` objects are used for internal data manipulation. If you are unfamiliar with `data.table` please see the [documentation](https://rdatatable.gitlab.io/data.table/index.html) and [cheatsheet](https://s3.amazonaws.com/assets.datacamp.com/img/blog/data+table+cheat+sheet.pdf). Prototype code may be written with other tools but will generally need be refactored to use `data.table` before submission (in PRs where help is needed with this please clearly state this).
```
pearsonca, line 53: I think there's also a worthwhile bullet here on *how* we write data.table.

I'm very inclined to bracket chaining approaches (performance-oriented). You seem to use one-liners with re-assignment (maintainability-oriented).

As much as I do not look forward to losing on this issue, I think you should pick one.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+
+```r
+library(data.table)
+
+no_dt_print <- function(dt) {
+    dt <- data.table::as.data.table(dt)
+    return(dt)
+}
+
+dt_print <- function(dt) {
```
pearsonca, line 69: if you use the pattern in the earlier code, can repeat it here.

I think its worth emphasizing that this is for both internal / external functions, for maintenance / debugging / consistency / etc reasons.

STYLE_GUIDE.md
pearsonca: also, to what extent does "importFrom" cause imports to be updated automatically?

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+
+This document outlines the style guide for the `epinowcast` package. This guide is a work in progress and will be updated as the package evolves. We welcome contributions to this guide and encourage you to raise issues or submit PRs if you have any suggestions.
+
+In addition to this guide we also follow the [tidyverse style guide](https://style.tidyverse.org/). This guide is a subset of the `tidyverse` style guide and outlines the additional style requirements for the `epinowcast` package.
+
+## Naming conventions
+
+- We use a `enw_` prefix to delineate functions that are part of the `epinowcast` package. This is to avoid conflicts with other packages and to make it clear to users which functions are part of the package.
+- The use of this prefix is not required for internal functions (i.e. functions that are not exported) or that are unlikely to have naming conflicts with other packages.
+
```
pearsonca, line 11: Feels like there are some other naming conventions here - e.g. `_add_` meaning that the submitted data.frame is modified? Perhaps ones that we want to, but do not yet, enforce.

pearsonca: any other vocabulary we want to define here? e.g. `rw` meaning random walk, `re` meaning...random effect?

pearsonca (changes requested):
Seems like a start, but needs some work?

I also think there's generally room to be aspirational here - have some rules we *want* to enforce, but that the code doesn't yet comply with. Helps us fix things as refactors go along, not introduce more work when we finally get around to it, etc.

STYLE_GUIDE.md
seabbs: Import from doesn't cause imports to be added to the DESCRIPTION but it does manage the namespace for you. 

Agree on the usage split for imports and suggests. Also worth saying the bar for a suggests is lower than for an import. We don't need to get to verbose here though as for most people who get this far this is likely to be common knowledge.

STYLE_GUIDE.md
seabbs: We do for everything that isn't an bracket operation (or at least we likely should). Can list for sure... how fun

STYLE_GUIDE.md
seabbs: I really like the guidance lining to as.data.table as it indicates to look there to see what it supports. It also means if they ever support more things then we don't need to redox

STYLE_GUIDE.md
seabbs: I shouldn't do anything. We can make a decision but I think that might need to be pushed to a new issue so others can weigh in. I agree on the relative trade off. Really I am waiting for the DT sugar to be released and then we can move to that with the base R pipe to get to some middle ground.

STYLE_GUIDE.md
seabbs: The add family could be flagged but it doesn't seem essential. I think pulling out the rw and re might be a bit detailed for this style guide?

seabbs:
Thanks Carl, some nice comments. 

I agree there is more we could cover but I think that needs to be discussed by a wider group first (so in a new issue, which you might like to write?) and potentially on a call if people feel strongly (sounds like a fun time to me). 

I'll make changes and then ping for another review.

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
...
+## Input types and checking
+
+- We support inputs that are coercible to `data.table` objects using `data.table::as.data.table()`. This includes `data.frame` and `tibble` objects. This should be clearly documented in the function documentation.
+- Any required inputs should be clearly documented in the function documentation.
+- We use an internal function `coerce_dt()` to check inputs are coercible to `data.table` objects and have the correct columns. This function is used in all functions that take data as input. The following function demonstrates this pattern:
+
+```r
+print_dt <- function(dt) {
+    dt <- epinowcast:::coerce_dt(dt, required_cols = c("date", "cases"))
```
Bisaloo, line 34: This example may give the wrong impression that you should write `epinowcast:::coerce_dt()` when we want just `coerce_dt()`

STYLE_GUIDE.md
```diff
@@ -0,0 +1,88 @@
+# Style Guide
+
+This document outlines the style guide for the `epinowcast` package. This guide is a work in progress and will be updated as the package evolves. We welcome contributions to this guide and encourage you to raise issues or submit PRs if you have any suggestions.
+
+In addition to this guide we also follow the [tidyverse style guide](https://style.tidyverse.org/). This guide is a subset of the `tidyverse` style guide and outlines the additional style requirements for the `epinowcast` package.
+
+## Naming conventions
+
+- We use a `enw_` prefix to delineate functions that are exported by the `epinowcast` package. This is to avoid conflicts with other packages and to make it clear to users which functions are part of the package.
```
Bisaloo, line 9: This should maybe be added to CI checks. The simplest option is likely to check the `NAMESPACE`

STYLE_GUIDE.md
seabbs: Yeah that is a good point. I had it without at first and then thought it won't work for people who aren't using a dev version of the package. Your probably right that is the least bad option though.

STYLE_GUIDE.md
seabbs: So that is a good idea but we haven't been quite consistent (mostly recently) so there are a few functions that don't do this (I guess we could right exceptions). There is an open issue to address the prefix more generally as when it was discussed at previous meetings the universal feeling was one of dislike.

STYLE_GUIDE.md
pearsonca: perhaps, was mostly proposing those as examples. how about the use of "date" as per the other thread?

my point here is that seems like there's a bit of nascent vocabulary and that it would be worthwhile to codify that *a bit* (obviously, can evolve as needed) so that we proceed intentionally with function / argument name choices.

STYLE_GUIDE.md
pearsonca: turn it into an issue then (adding the CI checks + renaming to get in compliance with them)?

seabbs: I don't think we want to. I think we want to discuss the prefix more generally first (which is an open issue).

STYLE_GUIDE.md
pearsonca: Eh, I think it's okay to point people at the chunk of code for the localized stuff (this happens somewhere for the linter, yeah?), rather than enumerate what we do and/or don't prefix with `data.table::`

STYLE_GUIDE.md
pearsonca: mmm, the preferred use of the pipe and lambda's in the code should definitely be a style guide item as well.

STYLE_GUIDE.md
seabbs: So I am just going to resolve this but happy to have an issue on this (needs to link to the current prefix discussion I think)

STYLE_GUIDE.md
seabbs: I think we need to discuss this in an issue/discussion. Could you take point on that? Resolving for this PR.

STYLE_GUIDE.md
seabbs: we don't want them used yet though as the new version of `data.table` hasn't been realised. Agree will need an issue to discuss when it is.

STYLE_GUIDE.md
seabbs: going to disagree for now (though I take your point)! Happy to have a wider discussion.

seabbs:
I think I have addressed the majority of both of your points. I agree there is more than can be done here on an aspirational level but quite a lot of that involves choices which really need to be made by the wider community (i.e. so we need issues to discuss and maybe bring up at a community meeting).

CONTRIBUTING.md
```diff
@@ -41,9 +44,9 @@ Please also follow these additional guidelines.
...
 * If possible, keep contributions small enough to easily review manually. It is okay to split up your work into multiple pull requests.
-* Format your code according to the [tidyverse style guide](https://style.tidyverse.org/). That formatting can achieved by running `style_pkg()` from [`styler`](https://github.com/r-lib/styler) (which rewrites the files) and `lint_package()` from [`lintr`](https://github.com/jimhester/lintr) (which provides a list of complaints for you to resolve). Note: `styler::style_pkg()` does not examine roxygen content i.e. `@examples` so you will need to check that manually.
+* Format your code according to our [package style](STYLE_GUIDE.md) or where not specified the [`tidyverse style` guide](https://style.tidyverse.org/). Automatic `tidyverse` styling can be achieved by running `style_pkg()` from [`styler`](https://github.com/r-lib/styler) (which rewrites the files) and `lint_package()` from [`lintr`](https://github.com/jimhester/lintr) (which provides a list of complaints for you to resolve). Note: `styler::style_pkg()` does not examine roxygen content i.e. `@examples` so you will need to check that manually.
```
pearsonca, line 47:
```suggestion
* Format your code according to our [package style](STYLE_GUIDE.md) or where not specified the [`tidyverse style` guide](https://style.tidyverse.org/). Automatic `tidyverse` styling can be achieved by running `styler::style_pkg()` from [`styler`](https://github.com/r-lib/styler) (which rewrites the files) and `lintr::lint_package()` from [`lintr`](https://github.com/jimhester/lintr) (which provides a list of complaints for you to resolve). Note: `styler::style_pkg()` does not examine roxygen content i.e. `@examples` so you will need to check that manually.
```

STYLE_GUIDE.md
```diff
@@ -0,0 +1,102 @@
...
+More generally when adding functions from external packages (i.e. even if they are already a dependency) the following should be followed:
+
+- Documented in function documentation using the `@importFrom` tag.
+- Used within functions using the `package::function` format (though we make exception for functions from `data.table` as these are all imported by `epinowcast`).
+
+## Input types and checking
+
+- We support inputs that are coercible to `data.table` objects using `data.table::as.data.table()`. This includes `data.frame` and `tibble` objects. This should be clearly documented in the function documentation.
+- Any required inputs should be clearly documented in the function documentation.
+- We use an internal function `coerce_dt()` to check inputs are coercible to `data.table` objects and have the correct columns. This function is used in all functions that take data as input. The following function demonstrates this pattern (note this requires the use of `devtools::load_all() in the package directory):
```
pearsonca, line 30:
```suggestion
- We use an internal function `coerce_dt()` to check inputs are coercible to `data.table` objects and have the correct columns. This function is used in all functions that take data as input. The following function demonstrates this pattern (note this requires the use of `devtools::load_all()` in the package directory):
```

STYLE_GUIDE.md
```diff
@@ -0,0 +1,102 @@
...
+dt[, cyl := cyl + 1]
+#over this
+dt_chain <- as.data.table(mtcars)[, mpg := mpg + 1][mpg > 20, cyl := 10][, cyl := cyl + 1]
+```
+
+- We also use `list` structures for more complex objects or where `data.table` is not appropriate. If the appropriate data structure is unclear for the problem at hand please flag this in the issue you are addressing or in the PR discussion.
+
+## Output types
+
+- For external functions we aim for the output to be a `data.table` object if possible unless a custom class is used (which we generally aim to inherit from the `data.table` class). This is to ensure consistency with the input types and to allow for easy chaining of functions. The following functions demonstrate this pattern:
```
pearsonca, line 72:
```suggestion
- For external functions we aim for the output to be a `data.table` object if possible unless a custom class is used (which we generally aim to inherit from the `data.table` class). This is to ensure consistency with the input types and to allow for easy chaining of functions.
```

pearsonca (approved):
Minor changes proposed; otherwise, seems like a good start.

pearsonca (approved):

PR #267, opened 2023-04-28
Issue 260: Refactoring the README

### LINKED ISSUE #260: Comments on `README.Rmd`
From @parksw3 

Probably a good idea to explain what is meant by day of reference vs report in this document too.

What is (1 | day)? Is this the random walk?

"Rather than using the methods supplied for epinowcast directly, package functions can also be used to extract nowcast posterior samples, summarise them, and then plot them" Is this sentence missing something? I'm confused what "package functions" are referring to. It sounds like you might be referring to a different package? Not sure why functions vs methods are different here...

Am curious why there is a consistent underestimation. Would be useful to give some information (e.g., what things you can try, what to read) for readers who want to learn more (but don't have to show how to actually do more complicated things)...

Original: https://github.com/epinowcast/epinowcast/issues/215#issuecomment-1519446636

seabbs:
## Description

This PR closes #260 by addressing @parksw3 comments on the package README. 

It substantially refactors the quick start with additional text and structure to explain areas of the model. One tension here is keeping the quick start relatively short. 

I have also updated the package citation and added @parksw3 as a contributor based on his comments.

## Checklist


README.Rmd
```diff
@@ -98,9 +98,9 @@ library(ggplot2)
...
+Nowcasting of right-truncated case counts involves the estimation of reporting delays for recently reported data. This requires data on these patterns for previous observations and typically this means the time series of data as reported on multiple consecutive days (in theory non-consecutive days could be used, but this is not yet supported in `epinowcast`). For this quick start, we use data sourced from the [Robert Koch Institute via the Germany Nowcasting hub](https://github.com/KITmetricslab/hospitalization-nowcast-hub/wiki/Truth-data#role-an-definition-of-the-seven-day-hospitalization-incidence). The data represent hospitalisation counts by date of positive test and date of test report in Germany up to October 1, 2021. For this, the originally weekly data were deconvolved to a daily resolution, and days with negative counts are adjusted.
```
adrian-lison, line 101: Maybe we can still improve on the definition of nowcasting here and simplify a little bit.

Maybe as the second sentence:
_For this, we need to count cases both by when they occurred (often called "reference date") and by when they were reported (often called "report date"). Then, the difference between the report date and the reference date is the reporting delay._

And do we really need the remark about the non-consecutive days here? I am worried this is only confusing people

README.Rmd
```diff
@@ -113,99 +113,146 @@ retro_nat_germany <- nat_germany_hosp |>
 retro_nat_germany
 ```
 
+This data is already in a format that can be used with `epinowcast`, as it contains
+
+- a reference date (column `reference_date`): the date of the observation, in this example the date of a positive test
+- a report date (column `report_date`): the date of report for a given set of observations by reference date
+- a count (column `confirm`): the total (i.e. cumulative) number of hospitalisations by reference date and report date
+
```
adrian-lison, line 121: I tried adding more structure that is easy to scan over

README.Rmd
```diff
@@ -113,99 +113,146 @@ retro_nat_germany <- nat_germany_hosp |>
...
+- A process (expectation) module that models the expected counts by date of reference (`reference_date`)
+- A parametric reference reporting model which models the reporting delay distribution from the date of reference
+- A non-parametric reporting model which models differences in the reporting delay distribution by date of report (`report_date`), for example, day-of-the-week effects on the reporting delay.
+
+In the following sections, we specify simple models for each of these modules. The appropriateness of these specifications will vary depending on your context. See our vignettes for further details on model specification and examples of more complex models.
+
+#### Process model
+
+A commonly used process model in nowcasting is to model the expected counts by date of reference via a lognormal random walk. This aims to strike a balance between capturing the underlying data generation process while being primarily dependent on the information provided by available observations and their reporting delays. Users may specify this model using the enw_expectation() function. Here, `day` refers to the number of days from the start of the data.
```
adrian-lison, line 159: Maybe rather geometric random walk? Or random walk on the log scale?

adrian-lison, line 159: I see what you are trying to say here (that this is a minimally informed smoothing prior and we thus give lots of influence to the data), but I find the current formulation a bit difficult to understand

README.Rmd
```diff
@@ -113,99 +113,146 @@ retro_nat_germany <- nat_germany_hosp |>
...
 
 ```{r}
-reference_module <- enw_reference(~1, distribution = "lognormal", data = pobs)
+expectation_module <- enw_expectation(
+  ~ 0 + (1 | day), data = pobs
+)
 ```
 
-Construct a model with a random effect for the date of report using the metadata produced by `enw_preprocess_data()`.
+Note that because the process model is parameterised as a growth rate, a random effect (i.e. `(1 | day)`) is equivalent to a random walk on the log scale. Here, we are defining a random effect as,
```
adrian-lison, line 167: _the process model is parameterised as a groth rate_ is not very clear/exact

README.Rmd
```diff
@@ -113,99 +113,146 @@ retro_nat_germany <- nat_germany_hosp |>
...
+
+$$ \text{day} \sim \mathcal{Normal}(0, \sigma) $$
+$$ \sigma \sim \mathcal{Half-Normal}(0, 1).$$
```
adrian-lison, line 170: Maybe rather just use \text?

README.Rmd
```diff
@@ -113,99 +113,146 @@ retro_nat_germany <- nat_germany_hosp |>
...
+
+Where $\mathcal{Half-Normal}$ is a normal distribution bounded to be greater than 0.
```
adrian-lison, line 172: Can't we assume this as common knowledge?

README.Rmd
```diff
@@ -113,99 +113,146 @@ retro_nat_germany <- nat_germany_hosp |>
...
-report_module <- enw_report(~ (1 | day_of_week), data = pobs)
+reference_module <- enw_reference(~1, distribution = "lognormal", data = pobs)
 ```
 
-Construct a model with a lognormal random walk on expected cases. See `enw_expectation()` for other suggested choices.
+Note that the default distribution is log-normal, hence the distribution argument could be omitted here.
+
+#### Reporting modifiers by report date
+
+Even where there is evidence that reporting processes can be approximated by a single distribution, there may be additional reporting modifiers that are not captured by the reference model. For example, reporting may be lower on weekends or holidays. We can specify a model for these modifiers using a hazard formulation (see [our model description for details](https://package.epinowcast.org/articles/model)) using the `enw_report()` function. Here we specify a model with a random effect for the day of the week to capture weekly seasonality in the reporting delay.
```
adrian-lison, line 186: Maybe it is better if we call these "effects" consistently, not modifiers?

README.Rmd
```diff
@@ -113,99 +113,146 @@ retro_nat_germany <- nat_germany_hosp |>
...
-  ~ 0 + (1 | day), data = pobs
-)
+report_module <- enw_report(~ (1 | day_of_week), data = pobs)
 ```
 
-### Model fitting
+### Precompiling the model
 
-First compile the model. This step can be left to `epinowcast` but here we want to use multiple cores per chain to speed up model fitting and so need to compile the model with this feature turned on.
+As `epinowcast` uses `cmdstan` to fit its models, it is necessary to first compile the model. This can be done using the `enw_model()` function. Note that this step can be left to `epinowcast`, but here we want to use multiple cores per chain to speed up model fitting and therefore compile the model with this feature turned on.
```
adrian-lison, line 194: Here it is always `cmdstan`, and later it is `cmdstanr`

adrian-lison (changes requested):
Important work, quick start section is a crucial part of the doc seen very early by the user

I think this is already much better, and have added a few more comments

adrian-lison:
Note that I added a commit without updating the md file, needs to be compiled in the end again

seabbs:
> I tried adding more structure that is easy to scan over 

Like it.

seabbs:
> Can't we assume this as common knowledge? 

I put this in in response to one of @parksw3 comments but agree it isn't helping here. Will drop (can revisit if  we think its needed).

README.Rmd
seabbs: Like it.

seabbs:
> quick start section is a crucial part of the doc seen very early by the user

I agree. I've made changes as suggested but I think this will likely take us at least a few more iterations to get right.

README.Rmd
```diff
@@ -219,7 +219,7 @@ nowcast <- epinowcast(data = pobs,
 
 ### The `epinowcast` object
 
-The `epinowcast()` function returns an `epinowcast` object which includes diagnostic information, the data used for fitting, and the underlying `cmdstanr` object.
+The `epinowcast()` function returns an `epinowcast` object which includes diagnostic information, the data used for fitting, and the underlying `cmdstan` object.
```
adrian-lison, line 222: I think cmdstanr calls this CmdStanModel https://mc-stan.org/cmdstanr/reference/CmdStanModel.html

README.Rmd
```diff
@@ -98,9 +98,9 @@ library(ggplot2)
...
+Nowcasting of right-truncated case counts involves the estimation of reporting delays for recently reported data. For this, we need to count cases both by when they occurred (often called "reference date") and by when they were reported (often called "report date"). Then, the difference between the reference date and the report date and is the reporting delay. For this quick start, we use data sourced from the [Robert Koch Institute via the Germany Nowcasting hub](https://github.com/KITmetricslab/hospitalization-nowcast-hub/wiki/Truth-data#role-an-definition-of-the-seven-day-hospitalization-incidence). The data represent hospitalisation counts by date of positive test and date of test report in Germany up to October 1, 2021.
```
adrian-lison, line 101: Minor typo, in "the difference between the reference date and the report date AND is the reporting delay", one and too much

adrian-lison (approved):
I think we are good to go for this release!

README.Rmd
seabbs: Good shout. I don't think we are doing this correctly in multiple places. Will make an issue.

Issue #273, opened 2023-05-08
Model initialisation issues due to a cmdstan update


seabbs:
It looks like there is a new patch to cmdstan that fixes the issues with input dimension. Unfortunately this looks to have flagged another few parameters with dimension problems causing model initialisation failures. 

This can be fixed without code changes by rolling back cmdstan versions but obviously this is not a sustainable solution. 

We should diagnose the dimension issues and then release a small patch release with this fix.

PR #280, opened 2023-05-10
0.2.2 release

seabbs:
As #273 means that the current release is broken with the latest `cmstan` release I think we should have a small patch release.

It doesn't really need to be reviewed I think as its so minor but do say if you have views (or think it should be reviewed). If no issues will update to a release version, merge this in, and make a GitHub release tomorrow.

Tagging @epinowcast/developers


seabbs (commented):
LGTM

PR #288, opened 2023-06-12
Issue 287: Improve PMF discretisation.

### LINKED ISSUE #287: Improve discretisation
Along with @parksw3 I have recently been looking at estimating continuous delays from data that is censored and right truncated. 

We have found that using a 1 day discretisation that starts from the observed delay gives biased results and reflects neither the mean or sd of the true distribution. 

One simple alternative we have explored that performs reasonably is to use a 2 day uniform window that is centred on the observed delay. Whilst this doesn't return the true censoring interval from combining primary and secondary events it does perform reasonably well. We should move to support this with the view to updating to a better approximation in the future.

seabbs:
## Description

This PR closes #287.

It does this by introducing a 2 day uniform censoring interval to the stan code. This comes at a small performance hit as we cannot optimise the transform to hazard quite as efficiently. This change has been tested by writing a simple simulator in R, and a version of the discretisation function for the PMF in R and then testing against these. Notable it still breaks down as an approximation at short delays and performs most poorly for delays of 0 and 1 days.

I have also updated the package documentation to reflect this change.

Note that the package example has had some performance reversion which I think is due to the change in how 0 delays are handled vs a bug. I think in general this is just better highlighting that the default model is not that robust. 

@parksw3 tagging for visibility. I'll update the results in the paper for the next version using this and also update the `epinowcast` model definition to cite our work once its in the public domain. 

## Checklist


seabbs:
Hmm not clear why benchmarking is not showing up here.

seabbs (commented):
LGTM (self-review). Code wise I am happy this is implemented as expected and that that implementation is documented as I'd expect. 

This leaves the conceptual issue of if this approximation makes sense which I think @parksw3 and I are fairly confident on. 

Going to leave this for another few days if anyone has time to take a look and if not will merge (with us being able to update in the future to improve the approximation)

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -42,54 +44,98 @@ vector adjust_lcdf_discrete(vector lcdf, int n, int max_strat) {
...
+    // normalize: cdf = cdf / (cdf[n] + cdf[n-1])
+    // both n and n-1 are in the denominator to account for 2 day width of the
+    // interval period
+    // If n were infinite this would be equivalent to dividing by 2.
+    if (n == 1) {
+      adjusted_lcdf = lcdf - lcdf[1];
+    }
+    if (n > 1) {
+      adjusted_lcdf = lcdf - log_sum_exp(lcdf[n], lcdf[n-1]);
+    }
```
adrian-lison, line 59: Nice!

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -42,54 +44,98 @@ vector adjust_lcdf_discrete(vector lcdf, int n, int max_strat) {
...
-  vector[n-1] lccdf;
-  // cccdf = 1 - cdf
-  lccdf = log1m_exp(lcdf[1:(n-1)]);
+  // h_0 = cdf_1
   lhaz[1] = lcdf[1];
-  // h = 1 - CCDF_n / CCDF_n-1
-  lhaz[2:(n-1)] = log1m_exp(lccdf[2:(n-1)] - lccdf[1:(n-2)]);
+  // h_n = p_n / (1 - sum^{n-1}_{d=0} p_d)
+  // h_n = cdf_n+1 - cdf_n-1 / (1 - sum^{n-1}_{d=0} cdf_d+1 - cdf_d-1)
+  // h_n = cdf_n+1 - cdf_n-1 / 1 - sum^n_{d=0} cdf_d + sum^n-2_{d=1} cdf_d
```
adrian-lison, line 91: Hmm, I would have written `1 - sum^n_{d=1} cdf_d + sum^n-2_{d=0} cdf_d` here

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -42,54 +44,98 @@ vector adjust_lcdf_discrete(vector lcdf, int n, int max_strat) {
...
+  // h_n = cdf_n+1 - cdf_n-1 / 1 - cdf_n - cdf_n-1
```
adrian-lison, line 92: You mean `1 - (cdf_n - cdf_n-1)`? But still I am missing a `-cdf_0` here based on the above step

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -42,54 +44,98 @@ vector adjust_lcdf_discrete(vector lcdf, int n, int max_strat) {
...
+  // h_n = (cdf_n+1 - cdf_n-1) / (cccdf_n - cdf_n-1)
```
adrian-lison, line 93: I am confused at this point... is `cccdf` the complementary cdf?

adrian-lison: One too many c's?

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -42,54 +44,98 @@ vector adjust_lcdf_discrete(vector lcdf, int n, int max_strat) {
...
+  // log(h_n) = log(cdf_n+1 - cdf_n-1) - log(cccdf_n - cdf_n-1)
+  if (n > 1) {
+    vector[n-2] lccdf;
+    // cccdf_n = 1 - cdf_n
+    lccdf = log1m_exp(lcdf[1:(n-2)]);
+    lhaz[2] = lprob[2] - lccdf[1];
+    if (n > 2) {
+      lhaz[3:(n-1)] = lprob[3:(n-1)] - log_diff_exp(lccdf[2:(n-2)], lcdf[1:(n-3)]);
+    }
+  }
```
adrian-lison, line 103: If the above derivation is correct, I agree!

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -42,54 +44,98 @@ vector adjust_lcdf_discrete(vector lcdf, int n, int max_strat) {
...
+vector log_hazard_to_logit_hazard(vector lhaz, int n) {
+  vector[n] logit_haz;
   // Logit transformation
-  lhaz[1:(n-1)] = lhaz[1:(n-1)] - log1m_exp(lhaz[1:(n-1)]);
+  logit_haz[1:(n-1)] = lhaz[1:(n-1)] - log1m_exp(lhaz[1:(n-1)]);
   // Set last logit transformed hazard to Inf (i.e h[n] = 1)
-  lhaz[n] = positive_infinity();
-  return(lhaz);
+  logit_haz[n] = positive_infinity();
+  return(logit_haz);
```
adrian-lison, line 115: good point

inst/stan/functions/discretised_logit_hazard.stan
```diff
@@ -42,54 +44,98 @@ vector adjust_lcdf_discrete(vector lcdf, int n, int max_strat) {
...
 }
 
 // Calculate discreteised logit hazard or log probability
 // up to a maximum oberved delay with a range of normalisation strategies
-// Calculate the daily hazard of reporting using parametric
-// distributions up to the maximum observed delay
+// Assumes that delays are doubly censored and that the interval width is
+// approoximated by a uniform.
 vector discretised_logit_hazard(real mu, real sigma, int n, int dist,
```
adrian-lison, line 122: small typos:
- approximated
- double-censored?

interval width is approximately uniformly distributed?

R/utils.R
```diff
@@ -71,8 +71,14 @@ convert_cmdstan_to_rstan <- function(functions) {
...
-    "lpmf[2:n] = log_diff_exp(lcdf[2:n], lcdf[1:(n-1)])",
-    "for (i in 2:n) lpmf[i] = log_diff_exp(lcdf[i], lcdf[i-1]);",
+    "lpmf[3:n] = log_diff_exp(lcdf[3:n], lcdf[1:(n-2)])",
+    "for (i in 3:n) lpmf[i] = log_diff_exp(lcdf[i], lcdf[i-2]);",
+    functions,
+    fixed = TRUE
+  )
+  functions <- gsub(
+    "lhaz[3:(n-1)] = lprob[3:(n-1)] - log_diff_exp(lccdf[2:(n-2)], lcdf[1:(n-3)]);", # nolint
+    "for (i in 3:(n-1)) lhaz[i] =  lprob[i] - log_diff_exp(lccdf[i-1], lcdf[i-2]);", # nolint
```
adrian-lison, line 81: _Welcome to stan, the eldorado for debugging_

tests/testthat/test-stan_discretised_logit_hazard.R
```diff
@@ -69,3 +69,93 @@ test_that("discretised_logit_hazard returns the same thing in both log
...
+  pmf <- pmf / cdf[n]
+  return(pmf)
+}
+
+# Simulate double censored data
+simulate_double_censored_pmf <- function(
+  alpha, beta, max, fun = rlnorm, n = 1000
+) {
+  primary <- runif(n, 0, 1)
+  secondary <- primary + fun(n, alpha, beta)
```
adrian-lison, line 99: Can you just help me here, why you have done it like this? I would have intuitively also expected a runif for `secondary`

adrian-lison (changes requested):
Nice implementation! I just have a couple of open questions regarding the theory/documentation, most likely due to my ignorance w.r.t double censoring...

adrian-lison:
A different thing I was wondering about: does the double-censoring assumption also have any implications for our hazard effects model?

vignettes/distributions.Rmd
```diff
@@ -26,13 +26,19 @@ The currently available parametric delay distributions are continuous probabilit
...
+$$p_{g,t,d} = \frac{F^{\mu_{g,t}, \upsilon_{g,t}}(d+1) - F^{\mu_{g,t}, \upsilon_{g,t}}(d-1)}{F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)}.$$
+
+Unless $d = 0$ then we instead have
+
+$$p_{g,t,1} = \frac{F^{\mu_{g,t}, \upsilon_{g,t}}(1)}{F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)}.$$
+
+Normalising by $F^{\mu_{g,t}, \upsilon_{g,t}}(D+1) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)$, ensures that the $p^{\prime}_{g,t,d}$ sum to 1. Since $F^{\mu_{g,t}, \upsilon_{g,t}}(D)$ is the probability of reporting before the maximum delay, this can also be interpreted as conditioning our distribution on the maximum delay.
+
+Note that because of the discretisation and normalization, the discrete delay distribution we obtain only approximates the original continuous distribution, and the approximation is worse for shorter delays.
```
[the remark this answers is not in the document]
seabbs: I think we do very specifically for `d=0` and `d =1` but more generally its also more of an issue for shorter delays as the magnitude of the bias vs other things is greater.

vignettes/distributions.Rmd
```diff
@@ -26,13 +26,19 @@ The currently available parametric delay distributions are continuous probabilit
...
 
-Note that because of the discretisation and normalization, the discrete random variable we obtain generally does not have exactly the same moments as the original, continuous distribution.
+More rigorously, we can define this in terms of the cumulative distribution function. Let $F^{\mu_{g,t}, \upsilon_{g,t}}$ be the cumulative distribution function of the continuous probability distribution with parameters $\mu_{g,t}$ and $\upsilon_{g,t}$. Then the probability of reporting a delay of $d$ days is
+$$p_{g,t,d} = \frac{F^{\mu_{g,t}, \upsilon_{g,t}}(d+1) - F^{\mu_{g,t}, \upsilon_{g,t}}(d-1)}{F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)}.$$
+
+Unless $d = 0$ then we instead have
+
+$$p_{g,t,1} = \frac{F^{\mu_{g,t}, \upsilon_{g,t}}(1)}{F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)}.$$
+
+Normalising by $F^{\mu_{g,t}, \upsilon_{g,t}}(D+1) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)$, ensures that the $p^{\prime}_{g,t,d}$ sum to 1. Since $F^{\mu_{g,t}, \upsilon_{g,t}}(D)$ is the probability of reporting before the maximum delay, this can also be interpreted as conditioning our distribution on the maximum delay.
```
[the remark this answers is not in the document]
seabbs: yes agree. If D was infinite we would need to divide by 2 (which is true as well I guess with the 1 day step but you obviously don't need to divide everything by one).

vignettes/distributions.Rmd
```diff
@@ -26,13 +26,19 @@ The currently available parametric delay distributions are continuous probabilit
...
 
-Note that because of the discretisation and normalization, the discrete random variable we obtain generally does not have exactly the same moments as the original, continuous distribution.
+More rigorously, we can define this in terms of the cumulative distribution function. Let $F^{\mu_{g,t}, \upsilon_{g,t}}$ be the cumulative distribution function of the continuous probability distribution with parameters $\mu_{g,t}$ and $\upsilon_{g,t}$. Then the probability of reporting a delay of $d$ days is
+$$p_{g,t,d} = \frac{F^{\mu_{g,t}, \upsilon_{g,t}}(d+1) - F^{\mu_{g,t}, \upsilon_{g,t}}(d-1)}{F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)}.$$
+
+Unless $d = 0$ then we instead have
+
+$$p_{g,t,1} = \frac{F^{\mu_{g,t}, \upsilon_{g,t}}(1)}{F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)}.$$
+
```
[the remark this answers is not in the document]
seabbs: I don't think this is true for `d=1`?  as the observation window is 0 - 2?

vignettes/distributions.Rmd
```diff
@@ -26,13 +26,19 @@ The currently available parametric delay distributions are continuous probabilit
...
 
-Note that because of the discretisation and normalization, the discrete random variable we obtain generally does not have exactly the same moments as the original, continuous distribution.
+More rigorously, we can define this in terms of the cumulative distribution function. Let $F^{\mu_{g,t}, \upsilon_{g,t}}$ be the cumulative distribution function of the continuous probability distribution with parameters $\mu_{g,t}$ and $\upsilon_{g,t}$. Then the probability of reporting a delay of $d$ days is
+$$p_{g,t,d} = \frac{F^{\mu_{g,t}, \upsilon_{g,t}}(d+1) - F^{\mu_{g,t}, \upsilon_{g,t}}(d-1)}{F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D)}.$$
```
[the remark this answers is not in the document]
seabbs: I'm not sure I totally follow though?

inst/stan/functions/discretised_logit_hazard.stan
seabbs: I think what is written is correct because the interval should be centred at the observed delay (for reasons discussed above). Given that context do you think what is written makes sense?

inst/stan/functions/discretised_logit_hazard.stan
seabbs: Shouldn't `cdf_0` always be zero?

inst/stan/functions/discretised_logit_hazard.stan
seabbs: ha yes sorry!

inst/stan/functions/discretised_logit_hazard.stan
seabbs: its a big if!

inst/stan/functions/discretised_logit_hazard.stan
seabbs: wow spelling Sam. Yes exactly. The combination of two uniforms is not a uniform so this is only an approximation.

tests/testthat/test-stan_discretised_logit_hazard.R
seabbs: Me as well which it turns out is wrong! 

So this is saying that `rprimary` can happen anywhere within a day. The secondary event then depends on this time and a draw from the probability distribution. 

The reason that the secondary event doesn't need a uniform is that this is captured in the observation process (the flooring). We then observe the delay which is the floored to a day version of the secondary event - the floored to a day version of primary event. 

Attached is a figure I have been playing around with to get at this. It shows the growth rate based real process compared two various approximations. The one that is equivalent to what we have here is the Uniform(-1, 1) one. 
[censoring-intervals.pdf](https://github.com/epinowcast/epinowcast/files/11789290/censoring-intervals.pdf)

seabbs:
> A different thing I was wondering about: does the double-censoring assumption also have any implications for our hazard effects model?

I think not as these are inherently discrete time but I could be wrong. The only reason this is a problem/something to correct is if someone wants to use the parametric delay distribution as a continuous distribution somewhere.

tests/testthat/test-stan_discretised_logit_hazard.R
seabbs: We really want the version that has a uniform on both primary and secondary but I don't have a good handle on what that looks like yet.

R/utils.R
seabbs: I tried to get the `cmdstanr` expose functions stuff setup so we can avoid this but it currently just segfaults on me 😿

vignettes/distributions.Rmd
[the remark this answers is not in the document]
adrian-lison: Ah no, I just mean the normalization. You wrote F^{\mu_{g,t}, \upsilon_{g,t}}(D + 1 ) + F^{\mu_{g,t}, \upsilon_{g,t}}(D) in the denominator, but I think it should be F^{\mu_{g,t}, \upsilon_{g,t}}(D) + F^{\mu_{g,t}, \upsilon_{g,t}}(D-1). This is also what you implemented, if I read your code correctly, and assuming that D means the longest delay here.

vignettes/distributions.Rmd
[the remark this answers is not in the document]
adrian-lison: Then this is again an indexing question? Because you write `p_{g,t,1}`. Shouldn't this then have:

F^{\mu_{g,t}, \upsilon_{g,t}}(2) - F^{\mu_{g,t}, \upsilon_{g,t}}(0)

in the nominator? And again, I think this is what you implemented.

So I think what you meant to write above is `p_{g,t,0} = ...`

vignettes/distributions.Rmd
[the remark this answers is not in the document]
adrian-lison: It would be awesome if we could at least conceptually separate these two parts from each other when explaining this in the documentation. Like, deriving the normalization factor for an infinite D first, and then adding the conditioning on D.

inst/stan/functions/discretised_logit_hazard.stan
adrian-lison: Yes, I was only critical of this intermediate step in line 91. I changed the notation slightly and added brackets. The result is corrrect and makes sense.

inst/stan/functions/discretised_logit_hazard.stan
adrian-lison: Did it myself and got the same result

seabbs:
Still sitting on this whilst we think about if it really is the best solution we can manage.

vignettes/distributions.Rmd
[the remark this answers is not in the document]
seabbs: because these have an overlap in the window with 0 and so the approximation is less accurate.

tests/testthat/test-stan_discretised_logit_hazard.R
seabbs: Does this make sense now @adrian-lison?

vignettes/distributions.Rmd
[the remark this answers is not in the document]
seabbs: yes, agree this should be indexed to 0.

vignettes/distributions.Rmd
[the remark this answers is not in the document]
seabbs: Do you think we can make this its own issue/PR pair to resolve/improve?

tests/testthat/test-stan_discretised_logit_hazard.R
seabbs: Happy with this as is?

vignettes/distributions.Rmd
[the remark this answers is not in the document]
seabbs: See 93325e3e

vignettes/distributions.Rmd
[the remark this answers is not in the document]
seabbs: So I have looked at the implementation again and I think the normalisation as written is correct and the issue is driven by the use of maximum delay as you note. I think this is the case because we are using $D$ here to mean the maximum observed delay in the model but due to interval censoring the actual window of possible delays we need to normalise over includes $D+1$ delays. 

In terms of phrasing... I think as written its okay but open to suggestions.

seabbs:
Rather than sitting on this I think we should merge is we can. @adrian-lison what do you think?

vignettes/distributions.Rmd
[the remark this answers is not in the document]
adrian-lison: Ah, I think now I get it. When you write

`adjusted_lcdf = lcdf - log_sum_exp(lcdf[n], lcdf[n-1]);` in your code `n=dmax=D+1`, so this is indeed what you implemented, right?

And the normalization is correct, because `sum_d=0^D (F_{d+1} - F_{d-1})` is indeed `F_{D+1} - F_{D}` (assuming F_d=0 for d<1)

adrian-lison:
> Rather than sitting on this I think we should merge as is if we can. @adrian-lison what do you think?

I agree. It will be a good improvement. I am also happy with all the documentation now (see my reply above, you can resolve the comment if my reasoning for my agreement with you makes sense).

LGTM

vignettes/distributions.Rmd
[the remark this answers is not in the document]
seabbs: Yes exactly, I think this does highlight that some improved commenting in the code is needed to make this clear (as it took me quite a while to work out what past sam was thinking. Will make an issue.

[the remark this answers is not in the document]
seabbs: Made an issue (#302) to do a better job commenting on what is happening.

seabbs:
Awesome will merge! I've made an issue to improve the documentation of the indexing in the Stan code as definitely needs to be easier to connect to the docs.

PR #295, opened 2023-08-01
Issue 294: Add check for missing grouping variables

### LINKED ISSUE #294: Cryptic bug in preprocessing when missing some grouping variables
**Describe the bug**
Cryptic error when preprocessing data which is stratfiied by more grouping variables than provided in the `by` argument.

**To Reproduce**
```
nat_germany_hosp <- epinowcast::germany_covid19_hosp[location == "DE"]
pobs <- enw_preprocess_data(nat_germany_hosp, max_delay = 30, by = NULL)

# Produces:
# Error in vecseq(f__, len__, if (allow.cartesian || notjoin || !anyDuplicated(f__,  : 
#  Join results in 632835 rows; more than 91791 = nrow(x)+nrow(i). Check for duplicate key values in i each of which join to the same group in x over and over again. If that's ok, try by=.EACHI to run j for each group to avoid the large allocation. If you are sure you wish to proceed, rerun with allow.cartesian=TRUE. Otherwise, please search for this error message in the FAQ, Wiki, Stack Overflow and data.table issue tracker for advice.
```

**Expected behavior**
This should either be handled gracefully or a clear error should be given.

adrian-lison:
## Description

This PR closes #294 by adding a clear error message when the data is stratified by more grouping variables than stated in the `by` argument of `enw_preprocess_data`. The user can then either add more variables to `by`, or aggregate the data by themselves beforehand. Maybe we should also offer a special preprocessing function for the latter in `epinowcast`?

The alternative would be to try to handle this gracefully, but I think we would end up chasing errors in the downstream preprocessing functions (the cryptic error occured in `add_max_reported`, which is quite early in the whole preprocessing pipeline, more errors would like arise after this was resolved). The above solution therefore seems cleanest to me. Alternative opinions welcome of course.

## Checklist


seabbs (changes requested):
Code implementation looks good. Need to fix some linting issues and add a test showing the edge case is fixed.

seabbs:
Nice stuff!

>  The user can then either add more variables to by, or aggregate the data by themselves beforehand. Maybe we should also offer a special preprocessing function for the latter in epinowcast?

Yes agree. 

> but I think we would end up chasing errors in the downstream preprocessing functions 

Agree

The test failure is coming from the fact that when we test we don't enforce "reference_date" or "report_date" to be present. We could either enforce this and change the tests as expected or (my preference) group on these variables only if they are present. If not documented then the fact we are doing this probably should be.

See: https://github.com/epinowcast/epinowcast/blob/49020b8a08ec9b38a45abb07a0fb28eaa62932e7/R/preprocess.R#L288

seabbs:
This felt like maybe it should be its own function vs a side effect when I started to look at the issues we were having with edge cases. What do you think?

I've implemented this hence why now passing checks. Happy to revert.

seabbs (approved):
LGTM. From my side, this looks good but I think given the changes I've made @adrian-lison you should do your own review. Happy for this to be merged as is or to do another round of review after more changes.

adrian-lison:
Looks good to me too. I made an issue (#304) for the aggregation function. Ready to merge. But how can I self-review?

seabbs (dismissed):
LGTM 🥳  (it just needs a review from not the person who made the PR I think and old reviews get dismissed when new commits come in)

seabbs (commented):
Somehow the check inside `enw_assign_group()` snuck back in. I have removed and so now all works as expected.

seabbs (approved):

seabbs (approved):

seabbs (approved):

PR #303, opened 2023-08-09
Issue 240: Flexible preprocessing and model timestep

### LINKED ISSUE #240: Support weekly data with a weekly process model
At the moment the package supports the use of weekly data with a daily process model (i.e reporting only happens once a week) though the performance of this model could be improved by allowing the specification of custom priors more readily (i.e. without stan code changes). 

However, for some use cases it may be desirable to have both weekly data and also a weekly process model. In particular, this would give a great speed-up in fitting times when delays are very long. This update would also be helpful as it would allow the package proprocessing to be more easily used for other tools such as `nowcaster` (@rafalopespx) if we standardise inputs at some point in the future.

As all preprocessing is date linked but the underlying model is index-linked this feature should be implementable by working through the preprocessing functions and adding a custom periodicity where required (as some time periods (i.e. months) are irregular a second option is likely needed to make sure that this is appropriately handled). This should be fairly straightforward, at least in theory.

To show this functionality is working at least an example script (i.e. `inst/examples`) would be helpful but ideally, a full vignette showing the difference between using a weekly process model and a daily one would be part of this package extension.

seabbs:
## Description

This PR closes #240. It adds a timestep argument to `enw_preprocess_data()`, the preprocessing functions it calls, and postprocessing functions so that the model can be used on timesteps other than a day. This results in a process model with a timestep of, for example, a week which may be helpful when delays are long, data is only available at this resolution and specifying a more exact model is not feasible or when computation is limited. Currently, this supports arbitrary numeric timesteps with syntax sugar for weekly data. 

This PR also adds a new data converter, `enw_aggregate_cumulative()`, which allows for the creation of data by a timestep from daily data. This is then used in a new example script to showcase a weekly timestep in use based on the package germany example data.

For further details of work done in this PR (and the outsstanding to dos) see this step by step list: https://github.com/epinowcast/epinowcast/issues/240#issuecomment-1669469756

For an example script that showcases the new functionaliy see: https://github.com/epinowcast/epinowcast/blob/seabbs/issue240/inst/examples/germany_weekly_process_model.R

Note that `plot`, `summary` etc can be used as normal to explore the fitted model.

## Checklist

Some people who I have discussed this with previously who may be interested in kicking the tires: @sbfnk @kgostic @zsusswein

@pearsonca or @adrian-lison if you have some capacity you could be good candidates as a reviewer of this PR.

@adrian-lison Unfortunately this may cause some pain as there are quite a few overlaps with #224 and #295. Happy to field those issues as they come up.


seabbs (commented):
LGTM (self-review). 

I've sight read back through all the code. Checked the new example, checked the README and run through the Germany vignette.

PR #351, opened 2023-10-27
Hotfix filter_reference_dates

adrian-lison:
## Description

This PR solves a bug that occurred when using `remove_days` in `filter_reference_dates` on data with missing reference dates, but that would only throw an error if using both `include_days` and `remove_days`.


R/preprocess.R
```diff
@@ -404,8 +404,7 @@ enw_filter_report_dates <- function(obs, latest_date, remove_days) {
 #'
 #' @param include_days if \code{earilest_date} is not given, the number
 #' of reference dates to include, ending with the latest reference
-#' date included once report dates have been removed. If specified
-#' this is indexed to `latest_date` or `remove_days`.
+#' date included (determined by `latest_date` or `remove_days`).
```
adrian-lison, line 407: I am not 100% sure what "once report dates have been removed" meant, but I think it was wrong. The report dates smaller than the earliest date were only removed in the end.

seabbs (commented):
Nice catch - looks good. Hopefully retriggered CI will all pass now.

seabbs (approved):
Nice catch - looks good. Hopefully retriggered CI will all pass now.

seabbs:
/document

I think the CI is just failing because of #360 and so this can be merged. Will keep an eye on the CI over the next few days and hot fix any problems

seabbs (approved):
LGTM. Ty ❤️

PR #365, opened 2023-11-15
Issue 344: Rename enw_delay_filter()

### LINKED ISSUE #344: enw_filter_delay function: Filter based on report - reference date
When using `epinowcast` it is often useful to compare nowcasts against later observed data. When doing this there are two sensible choices. The first is supported and is to filter the latest data at some future point and then use `enw_latest_data()`.  The second is to filter the data a consistent delay after the reference date so that all reference dates have the same horizon. This is currently unsupported. 

A new function called `enw_filter_delay` or similar that is based on the current `enw_filter_reference` and `enw_filter_report` functions and enables users to filter out reported data greater than some maximum delay. 

The new function should be clearly documented and have an example workflow. It should also have an informative news item that lays out why a user might want to use it.

kathsherratt:
## Description

This PR closes #344.

- Renames enw_delay_filter() to enw_filter_delay()
- Deprecated enw_delay_filter() with hard stop error (though this might be too harsh... I later noticed previous epinowcast depreciations used deprecate_warn() )

## Checklist


NEWS.md
```diff
@@ -51,6 +51,9 @@ This release is in development and not yet ready for production use.
 - Updated the README to include the non-parametric reference date model as an option and also added a new example showing how to use this model. See #313 by @seabbs and self-reviewed.
 - Added a new example showcasing how to fit a model to data reported weekly with a 3 day delay until any reports are non-zero with a weekly process model and a mixture of a parametric and non-parametric reference date model. See #348 by @seabbs and self-reviewed.
 
+## Depreciation
+- `enw_delay_filter()`: Deprecated with an error in favour of `enw_filter_delay()`. This renaming is to better reflect the function's purpose: see #344.
```
seabbs, line 55: can you add yourself as the PR author and me as the review (@seabbs).

seabbs: Potentially the title should be Depreciations?

R/preprocess.R
```diff
@@ -492,6 +492,27 @@ enw_latest_data <- function(obs) {
...
+#' @description `r lifecycle::badge("deprecated")`
+#' @return A `data.frame` filtered so that dates by report are less than or
+#' equal the reference date plus the maximum delay.
+#'
+#' @inheritParams enw_add_incidence
+#' @inheritParams enw_preprocess_data
+#' @importFrom lifecycle deprecate_stop
+#' @family preprocess
+#' @export
+enw_filter_delay <- function(obs, max_delay, timestep = "day") {
```
seabbs, line 506: I think we might want to stick with using `deprecate_warn` for now. Unless you feel strongly?

R/preprocess.R
```diff
@@ -492,6 +492,27 @@ enw_latest_data <- function(obs) {
...
+  lifecycle::deprecate_stop(
+    when = "0.2.3",
+    what = "enw_delay_filter()",
+    with = "enw_filter_delay()",
+    details = "Please file an issue if deprecating this \
+      function has caused any issues."
+  )
+}
+
 #' Filter observations to restrict the maximum reporting delay
```
seabbs, line 516: Could you have a crack at rewording this to better fit what the function does? It currently doesn't really match the name when I read it?

seabbs (changes requested):
Thanks @kathsherratt! This is great. I think we probably want to stick with deprecate_warn here unless you think strongly? 

Once you've had a chance to think about my comments could you ping me for another review?

Could you also add yourself to the package DESCRIPTION as a contributor?

R/preprocess.R
kathsherratt: Changed it a bit but not sure I got this right tbh

kathsherratt:
@seabbs pinging for a second review

seabbs (approved):
LGTM. Thanks a lot Kath!

NEWS.md
```diff
@@ -51,6 +51,9 @@ This release is in development and not yet ready for production use.
 - Updated the README to include the non-parametric reference date model as an option and also added a new example showing how to use this model. See #313 by @seabbs and self-reviewed.
 - Added a new example showcasing how to fit a model to data reported weekly with a 3 day delay until any reports are non-zero with a weekly process model and a mixture of a parametric and non-parametric reference date model. See #348 by @seabbs and self-reviewed.
 
+## Depreciations
+- `enw_delay_filter()`: Deprecated with an error in favour of `enw_filter_delay()`. This renaming is to better reflect the function's purpose. See #365 by @kathsherratt and reviewed by @seabbs.
```
jamesmbaazam, line 55: I believe this was changed to warn instead of error.

jamesmbaazam (commented):
Just noticed a minor mistake in the news item.

NEWS.md
seabbs: yes that is correct - good catch @jamesmbaazam! I guess we need a new issue and PR volunteer 🙈

jamesmbaazam: Sure. 😅

PR #376, opened 2023-11-18
Issue 188: Switch to rlang for warnings, messages and errors

### LINKED ISSUE #188: Switch to {cli} for all package messaging
_Update: See below that we are now thinking about using `cli` via `rlang`._

The `cli` package (https://cli.r-lib.org/) provides modern drop-in replacements for base R messaging/warning/errors as extensions of those in `rlang`. 

The package itself only has a single dependency (`utils`) and is a core part of Posit R infrastructure so receives good support. As with many Posit packages, this does mean breaking changes are potentially a concern but given the limited feature set we are going to make use of it is likely less of an issue and regardless these are generally well communicated. 

I'd suggest in the first instance taking a look at the `cli` package and discussing here if we think it's a good fit or not. Then working through our current messaging/warnings/errors and drop in replacements and finally perhaps give a lick of paint to our verbosity. 

From reading the package I think the drop in replacements we need are here: https://cli.r-lib.org/reference/cli_abort.html

nikosbosse:
## Description

This PR closes #188.

As discussed in #188 using rlang / cli has desirable properties for package messaging. This PR therefore changes all 
- instances of `stop()` to `rlang::abort()`
- instances of `warning()` to `rlang::warn()`
- instancse of `message()` to `rlang::inform()`

Remaining open questions: 
- If I'm not completely mistaken than `rlang` doesn't automatically use the `cli` package (as was discussed in #188). I think we'd also have to make the change suggested here: https://rlang.r-lib.org/reference/local_use_cli.html
- `check_modules_compatible()` used an additional argument to `warning()`: `immediate. = TRUE`. I left it in, but am not sure `rlang::warn()` will respect it. I currently lack the context to fully understand why the extra argument was there in the first place. If you think it's important this stays there I can investigate whether the functionality is respected by `rlang::warn()`. Otherwise the argument can be removed. 
- `enw_formula()` used an additional argument to `stop()`: `call. = FALSE`. I think the `rlang::abort` equivalent would be `call = NULL`. If that is desired I can make the change, otherwise we could just delete the argument as well and just rely on the default behaviour. 
- When building the package, a bunch of manual files were also updated. As far as I can tell those are only a few spaces here in there. The changes seemed unrelated to this PR, so I didn't include them here. ~~ Update: Apparently that was VSCode auto-styling files...

## Checklist


R/check.R
```diff
@@ -69,15 +76,16 @@ check_group_date_unique <- function(obs) {
...
   if (!"data" %in% names(module)) {
-    stop(
+    rlang::abort(
       "Must contain a list component specifying the data requirements for
        further modelling as a list"
     )
   }
   if (!is.list(module[["data"]])) {
-    stop(
+    rlang::abort(
```
pearsonca, line 88: My opinion: while in here, let's bring these up to style guide standards and say *what* must contain etc (e.g "Argument `module` ...") and similarly create a complete sentence, code element identified, etc for the message below.

*Not* a call for full review of these, just minor see-something, fix-something as we're doing all the error messaging.

Also need to update the STYLE_GUIDE.md to reflect usage of rlang.

R/check.R
```diff
@@ -91,19 +99,22 @@ check_module <- function(module) {
...
-      "number of threads per chain to the number of groups. To ",
-      "silence this warning, set the `likelihood_aggregation` ",
-      "argument in `enw_fit_opts` to 'groups'.",
+    rlang::warn(
+      paste0(
+        "Incompatible model specification: A missingness model has ",
+        "been specified but likelihood aggregation is specified as ",
+        "by snapshot. Switching to likelihood aggregation by group.",
+        " This has no effect on the nowcast but limits the ",
+        "number of threads per chain to the number of groups. To ",
```
pearsonca, line 114: Prefer to line break on sentences, rather than maximize characters used.

R/check.R
```diff
@@ -91,19 +99,22 @@ check_module <- function(module) {
...
-      "Incompatible model specification: A missingness model has ",
-      "been specified but likelihood aggregation is specified as ",
-      "by snapshot. Switching to likelihood aggregation by group.",
-      " This has no effect on the nowcast but limits the ",
-      "number of threads per chain to the number of groups. To ",
-      "silence this warning, set the `likelihood_aggregation` ",
-      "argument in `enw_fit_opts` to 'groups'.",
+    rlang::warn(
+      paste0(
+        "Incompatible model specification: A missingness model has ",
```
pearsonca, line 110: Again, I think we should clarify a bit here what precisely is causing the error, since the logic is automatically in the output message.

R/check.R
```diff
@@ -152,17 +163,17 @@ check_modules_compatible <- function(modules) {
...
-  forbidden_cols = NULL, group = FALSE,
-  dates = FALSE, copy = TRUE,
-  msg_required = "The following columns are required: ",
-  msg_forbidden = "The following columns are forbidden: "
-) {
+    data, select = NULL, required_cols = select,
+    forbidden_cols = NULL, group = FALSE,
+    dates = FALSE, copy = TRUE,
+    msg_required = "The following columns are required: ",
+    msg_forbidden = "The following columns are forbidden: ") {
```
pearsonca, line 173: I like this style `...) {` - but I thought the linter complains about it?

R/check.R
```diff
@@ -173,60 +184,65 @@ coerce_dt <- function(
     }
   }
 
-  if ((length(required_cols) > 0)) {     # if we have required columns ...
-    if (!is.character(required_cols)) {  # ... check they are check-able
-      stop("`required_cols` must be a character vector")
+  if ((length(required_cols) > 0)) { # if we have required columns ...
+    if (!is.character(required_cols)) { # ... check they are check-able
+      rlang::abort("`required_cols` must be a character vector")
```
pearsonca, line 189: Seems reasonable to add test that complains when required_cols isn't a character vector, to satisfy code coverage bot.

R/check.R
```diff
@@ -173,60 +184,65 @@ coerce_dt <- function(
...
+        )
       )
     }
   }
 
-  if ((length(forbidden_cols) > 0)) {    # if we have forbidden columns ...
+  if ((length(forbidden_cols) > 0)) { # if we have forbidden columns ...
     if (!is.character(forbidden_cols)) { # ... check they are check-able
-      stop("`forbidden_cols` must be a character vector")
+      rlang::abort("`forbidden_cols` must be a character vector")
```
pearsonca, line 209: ibid. ... but maybe just make me fix it by implementing coerce_dt

R/check.R
```diff
@@ -248,26 +264,30 @@ coerce_dt <- function(
...
+      )
     )
   }
 
   if (all_sequential_dates) {
     return(invisible(NULL))
   } else {
     if (exact) {
-      stop(
+      rlang::abort(
```
pearsonca, line 286: missing paste?

R/check.R
```diff
@@ -285,21 +305,26 @@ check_calendar_timestep <- function(dates, date_var, exact = TRUE) {
...
+#' @importFrom rlang abort
 check_numeric_timestep <- function(dates, date_var, timestep, exact = TRUE) {
   diffs <- as.numeric(
     difftime(dates[-1], dates[-length(dates)], units = "days")
   )
 
   if (any(diffs == 0)) {
-    stop(
-      date_var, " has a duplicate date. Please remove duplicate dates."
+    rlang::abort(
```
pearsonca, line 315: Two related notes:
 - seems like doing abort => paste0 in some places, abort => paste in others. Need to pick one to minimize future developer surprise.
 - also seems like several abort => pastes - DRY opportunity?

R/check.R
```diff
@@ -341,11 +368,12 @@ check_numeric_timestep <- function(dates, date_var, timestep, exact = TRUE) {
 #' @return This function is used for its side effect of stopping if the check
 #' fails. If the check passes, the function returns invisibly.
 #' @family check
+#' @importFrom rlang abort
 check_timestep <- function(obs, date_var, timestep = "day", exact = TRUE,
                            check_nrow = TRUE) {
   obs <- coerce_dt(obs, required_cols = date_var, copy = FALSE)
   if (!is.Date(obs[[date_var]])) {
-    stop(date_var, " must be of class Date")
+    rlang::abort(paste0(date_var, " must be of class Date"))
```
pearsonca, line 376: Probably wants a slgihtly better message here. Maybe

`sprintf("Column `%s` must be class `Date`.", date_var)`

?

R/check.R
```diff
@@ -354,7 +382,7 @@ check_timestep <- function(obs, date_var, timestep = "day", exact = TRUE,
 
   if (length(dates) <= 1) {
     if (check_nrow) {
-      stop("There must be at least two observations")
+      rlang::abort("There must be at least two observations")
```
pearsonca, line 385: ...of what?

R/data-converters.R
```diff
@@ -164,28 +169,35 @@ enw_linelist_to_incidence <- function(linelist,
...
   if (max_delay < obs_delay) {
-    message(
-      "Using the maximum observed delay of ", obs_delay,
-      " days as greater than the maximum specified to complete the incidence ",
-      "data.")
-       max_delay <- obs_delay
+    rlang::inform(
+      paste0(
+        "Using the maximum observed delay of ", obs_delay,
+        " days as greater than the maximum specified to complete the incidence",
```
pearsonca, line 190: max specified, which is ...?

R/formula-tools.R
```diff
@@ -202,7 +203,7 @@ remove_rw_terms <- function(formula) {
 #' epinowcast:::parse_formula(~ 1 + (1 | location) + rw(week, location))
 parse_formula <- function(formula) {
   if (!inherits(formula, "formula")) {
-    stop("'formula' must be a formula object.")
+    rlang::abort("'formula' must be a formula object.")
```
pearsonca, line 206:
```suggestion
    rlang::abort("`formula` must be a formula object.")
```

R/formula-tools.R
```diff
@@ -247,7 +249,7 @@ parse_formula <- function(formula) {
 rw <- function(time, by, type = c("independent", "dependent")) {
   type <- match.arg(type)
   if (missing(time)) {
-    stop("time must be present")
+    rlang::abort("time must be present")
```
pearsonca, line 252: Aren't missing variables reported as such? I'm a bit wary of providing a custom error message that doesn't actually say anything more than the base behavior would capture.

R/formula-tools.R
```diff
@@ -294,18 +297,22 @@ rw <- function(time, by, type = c("independent", "dependent")) {
 #' epinowcast:::construct_rw(rw(week, day_of_week), data)
 construct_rw <- function(rw, data) {
   if (!inherits(rw, "enw_rw_term")) {
-    stop("rw must be a random walk term as constructed by rw")
+    rlang::abort("rw must be a random walk term as constructed by rw")
```
pearsonca, line 300:
```suggestion
    rlang::abort("Argument `rw` must be a random walk term as constructed by `epinowcast:::rw`")
```

or some such. Bit ambiguous currently.

R/formula-tools.R
```diff
@@ -422,7 +436,7 @@ re <- function(formula) {
 #' epinowcast:::construct_re(random_effect2, mtcars)
 construct_re <- function(re, data) {
   if (!inherits(re, "enw_re_term")) {
-    stop("re must be a random effect term as constructed by re")
+    rlang::abort("re must be a random effect term as constructed by re")
```
pearsonca, line 439:
```suggestion
    rlang::abort("Argument `re` must be a random effect term as constructed by `epinowcast:::re`")
```

... or similar.

R/formula-tools.R
```diff
@@ -512,41 +528,44 @@ construct_re <- function(re, data) {
...
-          } else {
-            effects <- enw_add_pooling_effect(
-              effects, var_name = k,
-              finder_fn = function(effect, pattern) {
-                grepl(pattern, effect) & !grepl(":", effect, fixed = TRUE)
-              },
-              pattern = k
-            )
-          }
+        if (terms_int[i]) {
```
pearsonca, line 531: unclear why this should be changing - doesn't actually appear to be, but I'd also rather not be doubling checking that it isn't.

R/formula-tools.R
```diff
@@ -512,41 +528,44 @@ construct_re <- function(re, data) {
...
+        }
       } else {
         if (terms_int[i]) {
           effects <- enw_add_pooling_effect(
             effects,
             var_name = paste(gsub(":", "__", k, fixed = TRUE), collapse = "__"),
             finder_fn = function(effect, pattern) {
               grepl(pattern[1], effect) & grepl(pattern[2], effect) &
-              grepl(pattern[3], effect)
+                grepl(pattern[3], effect)
```
pearsonca, line 561: Does the linter want a space here?

I'm okay with a bit of side tidying along with other commits, but again: if something in the CI isn't complaining about it, please don't tweak spacing, line wraps, etc to personal taste. I am also regularly guilty of this, so totally sympathesize, but its a bad habit when working with others.

R/model-module-helpers.R
```diff
@@ -120,11 +122,11 @@ latest_obs_as_matrix <- function(latest) {
 convolution_matrix <- function(dist, t, include_partial = FALSE) {
   if (is.list(dist)) {
     if (length(dist) != t) {
-      stop("dist must equal t or be the same for all t (i.e. length 1)")
+      rlang::abort("dist must equal t or be the same for all t (i.e. length 1)")
```
pearsonca, line 125:
```suggestion
      rlang::abort("`length(dist)` must equal `t` or be the same for all t (i.e. length 1)")
```
Bit confused by the second note here - how is that checked?

R/model-module-helpers.R
```diff
@@ -120,11 +122,11 @@ latest_obs_as_matrix <- function(latest) {
...
     }
     ldist <- lengths(dist)
     if (!all(ldist == ldist[1])) {
-      stop("dist must be the same length for all t")
+      rlang::abort("dist must be the same length for all t")
```
pearsonca, line 129: Is this how the previous is checked?

R/model-modules.R
```diff
@@ -326,20 +330,21 @@ enw_report <- function(non_parametric = ~0, structural = ~0, data) {
...
   if (as_string_formula(r) %in% "~0") {
-    stop("An expectation model formula for r must be specified")
+    rlang::abort("An expectation model formula for r must be specified")
   }
   if (as_string_formula(observation) %in% "~0") {
     observation <- ~1
   }
   if (sum(generation_time) != 1) {
-    stop("The generation time must sum to 1")
+    rlang::abort("The generation time must sum to 1")
```
pearsonca, line 347: The `generation_time` pmf ? ...or similar?

R/model-validation.R
```diff
@@ -54,7 +55,7 @@
 enw_score_nowcast <- function(nowcast, latest_obs, log = FALSE,
                               check = FALSE, round_to = 3, ...) {
   if (!requireNamespace("scoringutils")) {
-    stop("scoringutils is required for this function to work")
+    rlang::abort("scoringutils is required for this function to work")
```
pearsonca, line 58:
```suggestion
    rlang::abort("The package `scoringutils` is required for this function to work.")
```

R/preprocess.R
```diff
@@ -123,7 +124,7 @@ enw_add_metaobs_features <- function(metaobs,
   # localize and check metaobs input
   metaobs <- coerce_dt(metaobs, required_cols = datecol)
   if (!is.Date(metaobs[[datecol]])) {
-    stop(sprintf("metaobs column '%s' is not a Date.", datecol))
+    rlang::abort(sprintf("metaobs column '%s' is not a Date.", datecol))
```
pearsonca, line 127:
```suggestion
    rlang::abort(sprintf("metaobs column `%s` is not a `Date`.", datecol))
```

R/preprocess.R
```diff
@@ -25,8 +25,8 @@
 #' @export
 #' @examples
 #' obs <- data.frame(
-#'  reference_date = as.Date("2021-01-01"),
-#'  report_date = as.Date("2022-01-01"), x = 1:10
+#'   reference_date = as.Date("2021-01-01"),
```
pearsonca, line 28: @seabbs one e.g. of roxygen change. not sure if this is linter complaint or not.

R/model-design-tools.R
```diff
@@ -265,23 +265,27 @@ enw_one_hot_encode_feature <- function(metaobs, feature, contrasts = FALSE) {
 #' @family modeldesign
 #' @export
 #' @importFrom purrr map
+#' @importFrom rlang abort
 #' @examples
 #' metaobs <- data.frame(week = 1:2)
 #' enw_add_cumulative_membership(metaobs, "week")
 #'
-#' metaobs <- data.frame(week = 1:3, .group = c(1,1,2))
+#' metaobs <- data.frame(week = 1:3, .group = c(1, 1, 2))
```
pearsonca, line 273: @seabbs the commas example - i would expect linter to complain about this.

pearsonca (commented):
Generally seems fine.

Two notes:
 - I made a review comment about clarifying the various error etc language while in here (and suggestions in a place or few). In retrospect, that actually seems big enough item to warrant its own PR
 - There seem to be a lot of changes here that I wouldn't expect the linter to trigger on, though my mental model of it may be wrong. That said: would make more sense to do that as a just a fixing-linting-issues PR.

R/check.R
pearsonca: Per overall comment: perhaps warrants own PR. @seabbs - thoughts? if so, let's try to capture notes / suggestions I made over this PR in an issue / update to STYLE_GUIDE.

R/model-design-tools.R
seabbs: yes me as well but it doesn't appear to

R/preprocess.R
seabbs: @nikosbosse ?

R/model-modules.R
seabbs: yes but not for this PR

R/check.R
seabbs: Yes agree but not for this PR.

R/check.R
seabbs: I think we should capture notes here and make suggestions in a new issue vs addressing here.

nikosbosse:
so it seems like VSCode decided to do some automatic changes on save to the files, resulting in lots of clutter - apologies for that. How should we proceed with these changes?

R/model-design-tools.R
nikosbosse: this seems to be the result of an automatic action of my VSCode linter (I'm normally using Rstudio, but it's currently unstable so switched to VSCode). Are you happy to keep these in? Or should I try and remove them again?

R/preprocess.R
nikosbosse: Apparently some auto-linting done by my VSCode 🙈

R/check.R
nikosbosse: Then probably `paste0` is the more general solution

nikosbosse: I also like the helper function - the `...` behaviour of the base `stop()` is quite nice

R/formula-tools.R
nikosbosse: Aaah I think this was my auto-linter changing the indentation... Can revert

R/formula-tools.R
nikosbosse: Not done on purpose!

NEWS.md
```diff
@@ -33,7 +33,7 @@ This release is in development and not yet ready for production use.
...
 - Added support to `enw_complete_dates()` for `min_date` and `max_date` arguments. These arguments allow users to specify the minimum and maximum dates to include in the output. This may be useful to users who want to ensure that their data is complete for a specified time period. See #340 by @seabbs and reviewed by @natemcintosh.
 - Added a new helper function `enw_one_hot_encode_feature()` for one hot encoding variables and binding them into the original data. This is useful when users want to include parts of variables in their models as binary indicators - for example giving a specific delay its own effect. See #348 by @seabbs and self-reviewed.
-
+- Switched to `rlang::abort()`, `rlang::warn()`, and `rlang::inform()` for all package messaging. See #188 by @nikosbosse.
```
pearsonca, line 36:
```suggestion
- Switched to `rlang::abort()`, `rlang::warn()`, and `rlang::inform()` for all package messaging. See #188 by @nikosbosse; reviewed by @pearsonca and @seabbs.
```

R/model-module-helpers.R
pearsonca: rec add to the generic error/warning review.

pearsonca (commented):
More minor notes, but I think the big thing to do here is revert the VScode IDE automated changes (with the exception of items that are otherwise linting, like the commas in the one roxygen example).

R/check.R
seabbs: if this is for another issue can it turned into one and this resolved?

R/check.R
seabbs: > ) and similarly create a complete sentence, code element identified, etc for the message below.

Can we make a new issue for this. 

> Also need to update the STYLE_GUIDE.md to reflect usage of rlang.|

I think this needs to be in this PR.

R/check.R
```diff
@@ -12,10 +12,11 @@
 check_quantiles <- function(posterior, req_probs = c(0.5, 0.95, 0.2, 0.8)) {
   stopifnot(
```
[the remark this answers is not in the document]
seabbs: Is there a `stopifnot` for `rlang` or does this require switching to if () {abort()}?

NEWS.md
```diff
@@ -33,7 +33,7 @@ This release is in development and not yet ready for production use.
...
 - Added support to `enw_complete_dates()` for `min_date` and `max_date` arguments. These arguments allow users to specify the minimum and maximum dates to include in the output. This may be useful to users who want to ensure that their data is complete for a specified time period. See #340 by @seabbs and reviewed by @natemcintosh.
 - Added a new helper function `enw_one_hot_encode_feature()` for one hot encoding variables and binding them into the original data. This is useful when users want to include parts of variables in their models as binary indicators - for example giving a specific delay its own effect. See #348 by @seabbs and self-reviewed.
-
+- Switched to `rlang::abort()`, `rlang::warn()`, and `rlang::inform()` for all package messaging. See #188 by @nikosbosse; reviewed by @pearsonca and @seabbs.
```
seabbs, line 36: This needs a new line below it, to be more informative as to why, to link to the PR and not the issue and to use the phrasing See #PR by @nikosbosse and reviewed by @pearsonca and @seabbs (to match current practice).

R/check.R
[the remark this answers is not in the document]
pearsonca: not really, see e.g.: https://github.com/r-lib/rlang/issues/1111

seabbs (changes requested):
Thanks for this @nikosbosse and thanks for the review @pearsonca. 

@nikosbosse I think the main things to do here are revert all the changes that aren't related to this PR and to deal with stopifnot cases. You will also need to rebase on `main`.

@pearsonca you brought up some good points about the actual content of the messages but I think those aren't really related to this PR. Can you please add some comments to the original (or a newer issue). 

On that point this is only a partial fix for the issue it targets as whilst it switches to rlang it doesn't improve the content of the messaging. Can we either make it clear this is a partial fix or open a new issue that is just for improving the content of the message/amount of messaging.

I have fixed all `lintr` issues on `main` so you will see some merge issues when you rebase. Alternatively it might be wise to start from `main` and revert this PR as it could be tricky to undo all the auto-formating.

_Looking again at the issue I see that the comments in it are what have crept the scope and that this does exactly address the feature request as written. Given that I think we need a new issue for improving message content and the perhaps another for improving the coverage/workflow of package messaging.

pearsonca:
I've resolved my comments that are now subject of #397 - still some open items, but those are mostly around the spacing issues (I also imagine catching this up to HEAD is going to be a bit of a pita).

nikosbosse:
Working on a new PR atm based on main - that's probably easier.

nikosbosse:
I replicated the changes made here on another branch. Should I just close this PR and make a new one? Or is there a different procedure you would prefer? 

 think the following points have not yet been addressed: 

> - If I'm not completely mistaken than rlang doesn't automatically use the cli package (as was discussed in https://github.com/epinowcast/epinowcast/issues/188). I think we'd also have to make the change suggested here: https://rlang.r-lib.org/reference/local_use_cli.html
> - check_modules_compatible() used an additional argument to warning(): immediate. = TRUE. I left it in, but am not sure rlang::warn() will respect it. I currently lack the context to fully understand why the extra argument was there in the first place. If you think it's important this stays there I can investigate whether the functionality is respected by rlang::warn(). Otherwise the argument can be removed.
> - enw_formula() used an additional argument to stop(): call. = FALSE. I think the rlang::abort equivalent would be call = NULL. If that is desired I can make the change, otherwise we could just delete the argument as well and just rely on the default behaviour.

What do you think?

seabbs:
> I replicated the changes made here on another branch. Should I just close this PR and make a new one?

Yes make a new PR!

PR #375, opened 2023-11-17
Issue 362: Move README quick start to vignette and improve general README formatting

pearsonca:
## Description

This PR closes #362 .

It splits the README into a tighter, reorganized version and a vignette (for the quickstart code).

## Checklist

- [X] My PR is based on a package issue and I have explicitly linked it.
- [X] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [X] I have read the [contribution guidelines](https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md).
- [X] I have tested my changes locally.
- [X] My code follows the established coding standards.
- [x] I have added a news item linked to this PR.
- [ ] I have updated the package development version by one increment in both `NEWS.md` and the `DESCRIPTION`.
- [x] I have reviewed CI checks for this PR and addressed them as far as I am able.

PR #390, opened 2023-11-22
Rely on pkgdown default behaviour to create 'Get started' page

Bisaloo:
## Description

Follow of #375 

This relies on pkgdown default behaviour of creating a "Get started" page for the vignette whose name is matching the package name. I believe it's better than creating custom sections.

Custom sections may break on future pkgdown update while the standard behaviour should be future proof. Custom sections also require custom translations if we ever plan on having a multilingual documentation website.

However, I may be missing something here. Please let me know.

## Checklist

- [ ] I have read the [contribution guidelines](https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md).


seabbs (commented):
Thanks for this @Bisaloo and good to have this discussion. 

The idea here is that in the future we might want more than one entry point into the package and that this nowcasting one might not be the main one (because nowcasting is   only a small part of current functionality and as the package develops this will only get smaller). Instead we probably want an overview getting started vignette that links out to all these other imagined vignettes. 

On the other hand as it stands those don't exist and so your suggestion probably makes a lot of sense at least for now? 

What do you think based on that additional context? I'm very open to your suggestion so just need a little convincing!

Bisaloo:
Thanks for the extra info and it does make sense. I don't have super strong feelings either :shrug: but a couple of extra thoughts:

> The idea here is that in the future we might want more than one entry point into the package and that this nowcasting one might not be the main one (because nowcasting is only a small part of current functionality and as the package develops this will only get smaller). Instead we probably want an overview getting started vignette that links out to all these other imagined vignettes.

When we arrive at this stage, I would suggest moving documentation to an online book format in a separate repo. At that point, I believe it would also be desirable to have split the package into several subunits to keep it more easily maintainable. So, having a longer documentation document outside of the pkgdown website would make sense.

> On the other hand as it stands those don't exist and so your suggestion probably makes a lot of sense at least for now?

Yes, as it is, the justification feels a little bit like "speculative generality", as described in [this nice blog post](https://chainsawriot.com/postmannheim/2023/09/29/mummies.html).

pearsonca:
I think this assertion is overconfident:
 
> Custom sections may break on future pkgdown update while the standard behaviour should be future proof. Custom sections also require custom translations if we ever plan on having a multilingual documentation website.

Both represent about the same level of declared `pkgdown` interface. As I practical matter, I'd tend to agree that pkgdown is more likely to continue doing *something* with a-name-matched vignette than maintaining the same spec for generating other stuff.

But they both seem pretty likely to me, and as such the "speculative" activity here is bothering with this change.

seabbs:
> the justification feels a little bit like "speculative generality",

Yes, it does. I think the counterargument is what will be the cost of changing it again in the future if we know where we are headed and we are headed there soon. Thanks for the blog post I enjoyed it.

That being said I think this only has a small cost so happy to go with the simple option for now. 

> I would suggest moving documentation to an online book format in a separate repo. At that point, I believe it would also be desirable to split the package into several subunits to keep it more easily maintainable.

Yes, I agree on both points. My hesitancy with splitting the package in the future is mainly working out how to manage multiple interrelated dependencies and how much overhead this might add vs the benefits of making it easier to maintain. I think its definitely something we will need to consider once pre and postprocessing has started to get more fleshed out and `epidist` is spun out of the work I have been doing with @parksw3.

 > I think this assertion is overconfident:
 
 I agree its perhaps a little more nuanced but then nuance is hard to write down without discussion. I definitely agree with @bisaloo that having a vignette for the `epinowcast.Rmd` entry point is likely to be a clear good starting point for many people?

pearsonca:
> Yes, it does. I think the counterargument is what will be the cost of changing it again in the future if we know where we are headed and we are headed there soon. Thanks for the blog post I enjoyed it.

I am however sympathetic to people looking generally for a vignette named after the package by default. I think the future evolution should correspond to a *different* getting started vignette anyway: one that gives a quick example of *each* of the package offerings, and signposts the getting-started-X sub vignettes which explore those offerings in a bit more detail (...and in turn point to the not-so-quick deep dives).

seabbs:
> I think the future evolution should correspond to a different getting started vignette anyway

Yes I agree. 


Okay the vibe I am getting is that we should accept this PR and also open an issue for making a true getting started vignette and pointing towards the nowcasting example and others in a more structured way?

_pkgdown.yml
```diff
@@ -20,17 +20,8 @@ development:
 authors:
   Sam Abbott:
     href: "https://www.samabbott.co.uk/"
-    
-navbar:
-  structure:
-    left: [reference, articles, news]
-    right: [github]
```
pearsonca, line 27: to confirm going to the default, which will be
```
navbar:
  structure:
    left: [intro, reference, articles, tutorials, news]
    right: [search, github]
```
left: no actual tutorials, but would auto-appear if added; otherwise same, except getting started article moved to intro
right: adding search

pearsonca (approved):
LGTM.

_pkgdown.yml
seabbs: yes I think so. Looks good to me.

seabbs (approved):
Agree looks good to me as well.

PR #394, opened 2023-11-23
DO NOT MERGE: Test PR for issue 385

seabbs:
…d code

 
## Description

Tests #386 using soft depreciated features and an old version of `cmdstan`.


seabbs:
Hmm @Bisaloo any ideas for getting the action to run? Otherwise just testing using `act` locally.

Bisaloo:
> Hmm @Bisaloo any ideas for getting the action to run? Otherwise just testing using act locally.

Maybe try removing the `branch` condition on your `push` trigger?

Bisaloo:
Could we target `main` to double check `R CMD check` is passing in these cases? Not 100% necessary but curious to confirm

seabbs:
Sigh these are all failing as they target the latest `cmdstan` version. Sigh!

seabbs:
😢 this version doesn't support multi-threading so fails... I think it proves the point though as tests running cmdstanr are only failing due to fitting issues and not due to the soft depreciations.

Issue #397, opened 2023-11-23
Standardize / clarify error / warning messages


pearsonca:
As highlighted during the review of #376, our warning / error / messaging content could use work.

Propose:
 - draft an update to STYLE_GUIDE.md covering the desired format of such messages AND where we should vs should _not_ provide them (e.g. because the base errors are sufficient and doing our own version is simply building future maintenance burden).
 - make that a draft PR
 - then comprehensive review of all warning / error / messages to meet the STYLE_GUIDE.md target.

Example issues with current approach:

## Calling Scope / Insufficient Feedback

```{r}
check_group_date_unique <- function(obs) {
  group_cols <- c("reference_date", "report_date", ".group")
  obs <- coerce_dt(obs, required_cols = group_cols, copy = FALSE)
  cells <- obs[, .(count = .N), by = group_cols]
  if (any(cells[, count > 1])) {
    rlang::abort(
      paste0(
        "The input data seems to be stratified by more variables ",
        "than specified via the `by` argument. Please provide additional ",
        "grouping variables to `by`, ",
        "or aggregate the observations beforehand."
      )
    )
  }
  return(invisible(NULL))
}
```

This message *might* be fine in practice, but it's a bit tough to tell without looking around at everywhere its used and trying it in situ. Perhaps this check needs to pass through the grouping columns as well? I'm generally in favor of explicitly showing the logic of how a requirement failed + particular values at failure.

## Explicitly Refer to Arguments

```{r}
# clipped from `check_module`
 if (!"data" %in% names(module)) {
    rlang::abort(
      "Must contain a list component specifying the data requirements for
       further modelling as a list"
    )
  }
  if (!is.list(module[["data"]])) {
    rlang::abort(
      "data must be a list of required data"
    )
  }
```

In general, we should explicitly refer to the argument with problems (in this , case `module` - though if we can get at the call stack and use whatever variable name the *user* passed to that argument, that would be ideal). If there are sub-elements that we're checking, we should be indicating the subelement connection (in this case, refer to `module[["data"]]` instead of "data").

PR #401, opened 2023-11-25
Issue 188: Switch to cli for warnings, messages and errors 

### LINKED ISSUE #188: Switch to {cli} for all package messaging

nikosbosse:
## Description

This PR closes #188.
It is also an updated version of #376, which was because since my VSCode decided to make lots of unrelated linting changes and main was updated in the meantime. This PR replicates the changes proposed in #376 but is based on an updated version of main and doesn't have the linting changes. 

As discussed in #188 using rlang / cli has desirable properties for package messaging. This PR therefore changes all 
- instances of `stop()` to `rlang::abort()`
- instances of `warning()` to `rlang::warn()`
- instancse of `message()` to `rlang::inform()`

Remaining open questions: 
- If I'm not completely mistaken than `rlang` doesn't automatically use the `cli` package (as was discussed in #188). I think we'd also have to make the change suggested here: https://rlang.r-lib.org/reference/local_use_cli.html
- `check_modules_compatible()` used an additional argument to `warning()`: `immediate. = TRUE`. I left it in, but am not sure `rlang::warn()` will respect it. I currently lack the context to fully understand why the extra argument was there in the first place. If you think it's important this stays there I can investigate whether the functionality is respected by `rlang::warn()`. Otherwise the argument can be removed. 
- `enw_formula()` used an additional argument to `stop()`: `call. = FALSE`. I think the `rlang::abort` equivalent would be `call = NULL`. If that is desired I can make the change, otherwise we could just delete the argument as well and just rely on the default behaviour. 

## Checklist


R/data-converters.R
```diff
@@ -354,7 +361,9 @@ enw_aggregate_cumulative <- function(
   obs, timestep = "day", by = NULL,
   min_reference_date = min(obs$reference_date, na.rm = TRUE), copy = TRUE
 ) {
-  stopifnot("The data already has a timestep of a day" = timestep != "day")
+  if (timestep == "day") {
+    abort("The data already has a timestep of a day")
```
pearsonca:
```suggestion
    rlang::abort("The data already has a timestep of a day")
```

R/formula-tools.R
```diff
@@ -294,18 +297,27 @@ rw <- function(time, by, type = c("independent", "dependent")) {
 #' epinowcast:::construct_rw(rw(week, day_of_week), data)
 construct_rw <- function(rw, data) {
   if (!inherits(rw, "enw_rw_term")) {
-    stop("rw must be a random walk term as constructed by rw")
+    rlang::abort(
+      paste(
```
pearsonca, line 301:
```suggestion
      paste0(
```
... also, any consideration of wrapper to give `stop(...)`s behavior?

R/formula-tools.R
```diff
@@ -318,15 +330,19 @@ construct_rw <- function(rw, data) {
   fdata <- data[, c(terms, rw$by), with = FALSE]
   if (!is.null(rw$by)) {
     if (is.null(fdata[[rw$by]])) {
-      stop(
-        "Requested grouping variable",
-        rw$by, " is not present in the supplied data"
+      rlang::abort(
+        paste0(
+          "Requested grouping variable",
```
pearsonca, line 335:
```suggestion
          "Requested grouping variable ",
```

pearsonca (commented):
Some minor items; otherwise good.

pearsonca (approved):

nikosbosse:
CI spell check complained about rlang, so I added an item to `inst/WORDLIST`. 

What should we do about the following items? 
> - If I'm not completely mistaken than `rlang` doesn't automatically use the `cli` package (as was discussed in #188). I think we'd also have to make the change suggested here: https://rlang.r-lib.org/reference/local_use_cli.html
> - `check_modules_compatible()` used an additional argument to `warning()`: `immediate. = TRUE`. I left it in, but am not sure `rlang::warn()` will respect it. I currently lack the context to fully understand why the extra argument was there in the first place. If you think it's important this stays there I can investigate whether the functionality is respected by `rlang::warn()`. Otherwise the argument can be removed. 
> - `enw_formula()` used an additional argument to `stop()`: `call. = FALSE`. I think the `rlang::abort` equivalent would be `call = NULL`. If that is desired I can make the change, otherwise we could just delete the argument as well and just rely on the default behaviour. 

Should I just remove the `call = NULL` and `immediate. = TRUE`?

seabbs:
[This](https://rlang.r-lib.org/reference/topic-condition-formatting.html) suggests to me we should be using `cli` directly. If we use `cli` in anyway we can use that to compose messages (as was the original intent) vs having to wrap abort around paste etc.

pearsonca:
> [This](https://rlang.r-lib.org/reference/topic-condition-formatting.html) suggests to me we should be using `cli` directly. If we use `cli` in anyway we can use that to compose messages (as was the original intent) vs having to wrap abort around paste etc.

Mmm, that seems like new issue territory. I think the `stop` => `abort` upgrade is a good first step (which provides useful reorganization in its own right), then next step is to go to `cli` (which will also provide a bit better interface, along the lines I poked at here, so need to address that notion *in this* PR).

pearsonca:
Hmm - on reflection and chat w/ @seabbs, actually I think going ahead w/ doing this in `cli` directly makes the most sense.

I also think what would be ideal here would be refactoring messages all into `glue`-formatted strings. We'd then have a very clear, consistent foundation to work from upgrading messages, expanding coverage, etc.

seabbs:
What do you think @nikosbosse? The reason to suggest going straight to `cli` is then we don't need to introduce the `paste` pattern into the code and can reformat once vs multiple times? It also seems like using `cli` directly is actually the recommended pattern vs via `rlang`?

nikosbosse:
One thing that I'm not completely able to oversee: 
As far as I can tell the `rlang` implementation has the advantage that it can use `cli` if it is available and otherwise fall back to base notifications. This would allow us to avoid an additional dependency on `cli`. Also commenters in #188 mentioned a few potential issues with `cli`. 

Otherwise I'm happy to follow your judgement call with this (and potentially copy what we do here in the `scoringutils` package)

seabbs:
> mplementation has the advantage that it can use cli if it is available and otherwise fall back

Is this actually true though? I'm not sure its the case? We don't want to implement a system that doesn't support within message pasting...

pearsonca:
> > mplementation has the advantage that it can use cli if it is available and otherwise fall back
> 
> Is this actually true though? I'm not sure its the case? We don't want to implement a system that doesn't support within message pasting...

... and re not adding a dependency, isn't using `rlang` adding a dependency as well? Perhaps a smaller mandatory dependency tree (aside: totally different conversation: there might be some NSE in rlang we want to use anyway), but still.

nikosbosse:
So current plan is to switch to completely switch to `cli`, right? I'l take a stab at this in the next days.

seabbs:
How are we getting along with this?

nikosbosse:
Work in progress!

seabbs:
Any blockers help would be useful for?

nikosbosse:
Is there a way I can trigger the CI checks myself?

One thing we'd need to discuss: As far as I understand, elements in a message vector sit on a new line. For example: 

image

Not sure we want this. Do we want to keep the calls to `paste0` in? Or just have line breaks?

nikosbosse:
ping ping :)

pearsonca:
> Is there a way I can trigger the CI checks myself?
> 
> One thing we'd need to discuss: As far as I understand, elements in a message vector sit on a new line. For example:
> 
> image
> Not sure we want this. Do we want to keep the calls to `paste0` in? Or just have line breaks?

...I increasingly dislike the "cleverness" in the tidyverse.

This is ... fine. Items should generally only be separated because our code needs line breaks, and that should roughly correspond with how long lines should be for users in the console.

We should *not* be managing inserting variables etc via chunks in the vector, but rather via glue formatting, so that also shouldn't be a source as undesired lines.

Lastly: however, we do need to document this gotcha in the style guide, and possibly also the templates for new features. Something like "does your new feature handle error conditions (e.g. bad arguments)? If so, have you verified via tests that enw raises these errors appropriately? Have you checked the style guide to ensure your error messages meet expectations?"

pearsonca:
> Lastly: however, we do need to document this gotcha in the style guide, and possibly also the templates for new features. Something like "does your new feature handle error conditions (e.g. bad arguments)? If so, have you verified via tests that enw raises these errors appropriately? Have you checked the style guide to ensure your error messages meet expectations?"

n.b. while I'd lean towards this last item now, @seabbs likely disagrees. Totally reasonable as a distinct, more manageable/compartmentalized follow-on issue / PR combo *if* raised as part of resolving this issue / PR.

seabbs:
How are getting on with this? Just noting that we have another PR (#402) that is stuck on this so it would be great to get it merged in. Can I help get it moving?

nikosbosse:
Working on this today. 

> I think we want to use glue functionality for this.

So essentially we replace calls to `paste0()` with calls to `glue()`, right?

pearsonca:
> So essentially we replace calls to `paste0()` with calls to `glue()`, right?

Might be able to just provide glue-formatted strings (and it will find things in the environment) - I vaguely recall something to that effect, but could be fever-dreaming it.

pearsonca:
> > So essentially we replace calls to `paste0()` with calls to `glue()`, right?
> 
> Might be able to just provide glue-formatted strings (and it will find things in the environment) - I vaguely recall something to that effect, but could be fever-dreaming it.

Yeah, cursory revisit to the documentation for cli + glue indicates that it pulls from the environment, so: should be able to just pass the format strings, and cli will take care of the glue-ing. Obviously, double check, but ideal would be to *not* have to explicitly invoke glue.

R/check.R
```diff
@@ -193,7 +193,7 @@ coerce_dt <- function(
           "{toString(required_cols[!(required_cols %in% colnames(dt))])}",
           "but are not present among ",
           "{toString(colnames(dt))}",
-          "(all `required_cols`: {toString(required_cols)})",
+          "(all `required_cols`: {toString(required_cols)})"
```
pearsonca, line 196:
```suggestion
          "(all {.arg required_cols}: {toString(required_cols)})"
```
... should be using the supported inline formatting for `cli`: https://cli.r-lib.org/reference/inline-markup.html

pearsonca (commented):
I think this is mostly fine for this pass; I suggested one change, and was going to suggest more in a similar vein, but I think we can leave purposeful use of the cli inline formatting to a separate issue / PR after we solve the style guide for errors issue / PR.

nikosbosse:
I think this is now ready for another pair of eyes, @pearsonca

DESCRIPTION
```diff
@@ -97,6 +97,7 @@ Imports:
     ggplot2,
     posterior,
     purrr,
+    rlang,
```
seabbs, line 100: don't you mean `cli`?

NEWS.md
```diff
@@ -137,7 +138,7 @@ Full details on the changes in this release can be found in the following sectio
...
 - Added tests for all internal `check_` functions used to check inputs. See #217 by @seabbs and reviewed by @adrian-lison.
 - Removed the problematic double specification of default arguments for `target_date` in `enw_metadata()` as flagged in #212 by @pearsonca using `formals()` to instead detect the default values from the function specification. See #232 by @seabbs and self-reviewed.
-- In the words of Jenny Bryan: "there is no else, there is only if." Having else after `return()` of `stop()` increases the number of branches in the code, which makes it harder to read. It also translates into a higher cyclomatic complexity. We have removed all else statements after `return()` and `stop()` in the package. See #229 by @Bisaloo and reviewed by @seabbs.
+- In the words of Jenny Bryan: "there is no else, there is only if." Having else after `return()` of `stop()` increases the number of branches in the code, which makes it harder to read. It also translates into a higher cyclomatic complexity. We have removed all else statements after `return()` and `cli::cli_abort()` in the package. See #229 by @Bisaloo and reviewed by @seabbs.
```
seabbs, line 141: Can we revert this as its from a previous version

R/check.R
```diff
@@ -7,13 +7,12 @@
...
-#'
+#' @importFrom cli cli_abort
 #' @family check
 check_quantiles <- function(posterior, req_probs = c(0.5, 0.95, 0.2, 0.8)) {
-  stopifnot(
-    "Please provide probabilities as numbers between 0 and 1." =
-    all(data.table::between(req_probs, 0, 1, incbounds = FALSE))
-  )
+  if (!all(between(req_probs, 0, 1, incbounds = FALSE))) {
+    cli::cli_abort("Please provide probabilities as numbers between 0 and 1.")
```
seabbs, line 14: NAMESPACE needs updating?

R/check.R
```diff
@@ -48,16 +47,20 @@ check_group <- function(obs) {
...
   obs <- coerce_dt(obs, required_cols = group_cols, copy = FALSE)
   cells <- obs[, .(count = .N), by = group_cols]
   if (any(cells[, count > 1])) {
-    stop("The input data seems to be stratified by more variables ",
-         "than specified via the `by` argument. Please provide additional ",
-         "grouping variables to `by`, ",
-         "or aggregate the observations beforehand.")
+    cli::cli_abort(
+      c(
+        "The input data seems to be stratified by more variables than",
```
seabbs, line 59: so will this print over multiple lines now or have the same behaviour as before.

R/check.R
```diff
@@ -68,18 +71,19 @@ check_group_date_unique <- function(obs) {
...
 #' @return NULL
 #'
+#' @importFrom cli cli_abort
 #' @family check
 check_module <- function(module) {
   if (!"data" %in% names(module)) {
-    stop(
-      "Must contain a list component specifying the data requirements for
-       further modelling as a list"
+    cli::cli_abort(
```
seabbs, line 78: as we aren't using package::function for the most part and `cli` functions are all prefixed with the package name do we really need to do so here?

R/data-converters.R
```diff
@@ -370,9 +378,14 @@ enw_aggregate_cumulative <- function(
   init_ref_date <- min_reference_date + internal_timestep - 1
   agg_obs <- obs[report_date >= init_ref_date]
 
-  stopifnot(
-    "There are no complete report dates (i.e. report_date >= min_date + timestep - 1)" = nrow(agg_obs) > 0 # nolint
-  )
+  if (nrow(agg_obs) == 0) {
+    cli::cli_abort(
+      c(
+        "There are no complete report dates",
```
seabbs, line 384: can this really not just be one line?

inst/WORDLIST
```diff
@@ -75,6 +75,7 @@ postprocess
 postprocessing
 predictives
 readibility
+rlang
```
seabbs, line 78: remove

seabbs (changes requested):
This is looking really good Nikos and is nearly there I think. Turned into a fair bit of work!

seabbs:
@nikosbosse you also need to add yourself as a contributor in the DESCRIPTION please!

R/check.R
pearsonca: print over multiple lines; i think we should triage messaging generally as part of the be-written issue for update to style guide for error messaging.

R/check.R
pearsonca: i'd be fine with not using `cli::cli_...` - I thought we had written down somewhere that we generally preferred the `package::function` syntax, but I think prefixed ones could be an exception (assuming we do even have that rule? not finding it).

R/check.R
```diff
@@ -175,36 +184,36 @@ coerce_dt <- function(
...
-        msg_required,
-        toString(required_cols[!(required_cols %in% colnames(dt))]),
-        " but are not present among ",
-        toString(colnames(dt)),
-        "\n(all `required_cols`: ",
-        toString(required_cols),
-        ")"
+      cli::cli_abort(
+        c(
+          "{msg_required}",
```
pearsonca, line 193:
```suggestion
          msg_required,
```
I think this wants to remain the variable, so that we can pass in a glue-format string here?

pearsonca: should double check the behavior of that, however.

R/check.R
```diff
@@ -175,36 +184,36 @@ coerce_dt <- function(
...
-        msg_forbidden,
-        toString(forbidden_cols[forbidden_cols %in% colnames(dt)]),
-        " but are present among ",
-        toString(colnames(dt)),
-        "\n(all `forbidden_cols`: ",
-        toString(forbidden_cols),
-        ")"
+      cli::cli_abort(
+        c(
+          "{msg_forbidden}",
```
pearsonca, line 211:
```suggestion
          msg_forbidden,
```
ibid.

R/data-converters.R
pearsonca: counter: the next line should probably give parsed values. So maybe more like
There are no rows with report_date >= min_date + timestep - 1
i.e. report_date >= {min_date + timestep - 1}

However: this is drifting into write-the-style-guidance-then-fix new PR territory.

R/utils.R
```diff
@@ -211,11 +212,10 @@ coerce_date <- function(dates) {
 
   if (anyNA(res)) {
     bads <- is.na(res)
-    stop(sprintf(
-      "Failed to parse with `as.IDate`: {%s} (indices {%s}).",
-      toString(dates[bads]),
-      toString(which(bads))
-    ))
+    cli::cli_abort(
```
pearsonca, line 215: That's a mildly annoying false positive on the linter. I guess mark w/ # no lint after bads?

R/utils.R
```diff
@@ -244,18 +245,22 @@ get_internal_timestep <- function(timestep) {
       day = 1,
       week = 7,
       month = "month",  # months are not a fixed number of days
-      stop(
-        "Invalid timestep. Acceptable string inputs are 'day', 'week',",
-        " 'month'."
+      cli::cli_abort(
```
pearsonca, line 248: not for this PR, but maybe should be refactoring this to work with match.arg

tests/testthat/test-check_group_date_unique.R
```diff
@@ -18,7 +18,7 @@ test_that("check_group_date_unique stops with duplicated groups", {
 
   # Expect error due to duplicated combination
   expect_error(check_group_date_unique(obs),
-               "The input data seems to be stratified by more variables than specified via the `by` argument. Please provide additional grouping variables to `by`, or aggregate the observations beforehand.")
+               "The input data seems to be stratified by more variables than\nspecified via the `by` argument. Please provide additional grouping\nvariables to `by`, or aggregate the observations beforehand.")
```
pearsonca, line 21: maybe should be adding a helper function to prune newlines and then compare that output - moving around \n s is going to get real tedious.

pearsonca (commented):
Some big picture thoughts:
 - my take on the importFrom on each function: that's repetitious and unnecessary. we should be doing that at the global level for these widely used functions. importFrom is more for special cases.
 - agree w/ @seabbs re just using `cli_...` rather than prefixing with `cli::`
 - the newlines in tests text comparison need to go - that's creating a maintenance burden. probably just apply a newline to nothing gsub on error messages.
 - there are several places to tweak with updated style guidance, but that's for another issue / PR (#397) - however, might be worth tagging some lines for attention during resolution of that.

nikosbosse:
@pearsonca so your suggestion is to remove all "\n" from the tests and instead use `gsub()` / shortening the relevant part of the error message?

pearsonca:
> @pearsonca so your suggestion is to remove all "\n" from the tests and instead use `gsub()` / shortening the relevant part of the error message?

Basically, yes - the line breaks are going to be a moving target as we refine error messages in the future, so shouldn't be giving ourselves known future maintenance burden where we can avoid it.

nikosbosse:
Hm I didn't find any styling insensitive string matcher - the only thing I found is that you can use additional arguments to `grepl()` in `expect_error()`. Also I realised that the helper function is maybe a bit more complicated - we'd have to catch the errors and then return them as a string + also wrap the string that we match the error against. 
Maybe this should be done in a separate PR? 

Iff we decide to do it that way, then maybe it doesn't make much sense to do a lot to the error messages right now? Otherwise I could for example just shorten the error messages to avoid the line breaks. What do you think?

seabbs:
So I think this had got a bit out of hand. 

I've rebased on `main`, reverted the test changes, and fallen back to wrapping with `paste0` to avoid all the very odd line breaks that had been introduced. 

This should make this PR essentially just dropping in new functions with few to no other changes. I suggest a follow up issue PR to improve aspects of messaging as we really aren't taking advantage of the nice `cli` functionality. 

Good job on this all - ended up being a bit more than expected!

seabbs (dismissed):
I think this is all essentially a swap in now? @nikosbosse happy with the updates?

nikosbosse:
Haven't looked at it in detail yet - maybe it's easier to use `glue()` instead of `paste0()`?

nikosbosse:
Otherwise all good from my side :)

seabbs:
In the interest of moving things along I am going to propose we merge this and have a follow up issue to deal with chasing things up? @pearsonca?

pearsonca (approved):
Yep, leave details to another round on this element (aside from fixing the one whitespace item).

seabbs (approved):

pearsonca (approved):

seabbs:
Woo! Good work on this all and thanks for the first contribution @nikosbosse!

adrian-lison:
Thanks for the work @nikosbosse and @seabbs!

I don't know much about linting - but would it be useful and possible to have a linter that complains if you accidentally use the base R messaging functionality in the future?

seabbs:
That is a great idea but I also don't have much of a clue how to do. Might be worth chucking up as an issue!

Issue #233, opened 2023-04-17
Refine input checking


seabbs:
Currently, the package uses a series of custom `check_` functions to check the inputs of user-facing functions. These need to be refined and iterated on so that they are more clearly implemented and cover the full range of inputs. 

There are quite a few approaches for doing this (with the simplest being the currently used `if` and `stop` approach. 

Some suggested alternatives (see [here](https://github.com/epinowcast/epinowcast/pull/229) for the original discussion:

## From @pearsonca 

For error-checking / messages, I think this generally reads more clearly (comments for emphasis - they're unnecessary imo, especially if this idiom is repeated through a library)

```r
stopifnot( # ... we're doing some error checking
  "checking this feature" = ..., # against some logical expression
  ... # repeat for all the desired checks
)
```

Aside: I am on-board with the don't-if-else-your-errors mentality, just not necessarily the return() bit. There are two useful idioms for returns in branching code:

- pre declare the object to manipulate, make the manipulations in the different branches, return after the branches.
- have each branch return at its end
I have a slight preference for the first, but its not always practical, nor is it always more readable. As such, disagree with making it a general principle / practice - that's for strong preferences, particularly where the lazy alternatives pose distinct downsides.

looking at the linked JB example, agree the LHS is code smell, but again: the `stopifnot(...)` idiom would be a much clearer solution than the RHS. That idiom telegraphs: "Here are all the things we can check up front and immediately exit, for reasons".

However, sometimes there's actually some branching logic (e.g. when embracing some polymorphism in function arguments, though yes that too can go overboard), and you can't properly check exit conditions until you've made some progress on evaluating arguments.

In terms of actual code, for check_date, I'd prefer something akin to:

```r
with(obs, stopifnot(
  "Both `reference_date` and `report_date` must be columns." =
     !(is.null(reference_date) && is.null(report_date)),
  "`reference_date` must be a column." = !is.null(reference_date),
  "`report_date` must be a column." = !is.null(report_date),
 local = ... # not quite sure here; seems like this is an internal function used for checks - parent.frame?
))
```

probably before the internal copy steps as well?

## From @Bisaloo 

I agree that `stopifnot()` is easier to understand for simple checking. I can do the change for the case at hand.

## From @TimTaylor

Unsolicited, 😬, but hopefully useful comments:

`stopifnot()` definitely looks nice.
[...]

PR #412, opened 2023-12-04
depend on coerceDT

pearsonca:
## Description

This PR address elements of #233 by consolidating to depend on `coerceDT` library (still a WIP, hence expecting this PR to be draft for a bit while we work both libraries).

## Checklist

- [X] My PR is based on a package issue and I have explicitly linked it.
- [X] I have read the [contribution guidelines](https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md).


pearsonca:
Looking at how we use `coerceDT` here and the related functionality in scoringutils, seems like there is appetite for "custom" messages.

I think a good 90-95% solution would be:
 - provide `crc_message` (coerce, for either select or drop warnings), `fbd_message`, and `exp_message` (forbid, expect) arguments
 - those arguments can either be a character string, which will be used as a format string, and passed the offending column names combined as a comma-separated string
 - ... or a function which receives the offending column names and returns a string (future: some of the function-based failing values / other info)

and where the default behavior is set by default values for those arguments.

@seabbs / @nikosbosse - thoughts?

pearsonca:
There are two epinowcast-specific items that don't really generalize (to coerceDT package): dates + group.

Propose:
 - for dates, have an epinowcast package option for required date column names, use that as an addition to the select (this may also require a new feature for coerceDT - a "." select support
 - for group, re-introduce a check_group function

pearsonca:
> There are two epinowcast-specific items that don't really generalize (to coerceDT package): dates + group.
> 
> Propose:
> 
> * for dates, have an epinowcast package option for required date column names, use that as an addition to the select (this may also require a new feature for coerceDT - a "." select support
> * for group, re-introduce a check_group function

Alternative-and-actually-general approach for `.group` - offer a `coalesce` argument on coerceDT (and therefore makeDT). Meaning "`select` this column if present, otherwise use this default value"

seabbs:
This all sounds and looks good but going to close until we have a version that is ready to PR in!

PR #416, opened 2023-12-14
#407 Add pre-compiled cmdstan models

### LINKED ISSUE #407: Add pre-compiled cmdstan models using instantiate
Currently, users need to compile our models when they first run them. In addition to this as we are trying to follow CRAN policy we also only save these compiled models to the temporary directory and therefore users may have to re-compile the model in new sessions (unless they manually set the target directory).

Instead of doing this we could use the new [instantiate](https://wlandau.github.io/instantiate/) package which enables compiling `CmdStan` models at installation and facilitates providing these to the user. Potentially this could also allow for binary packages to be built that include the stan model but I don't think this would work.

**Describe alternatives you've considered**
There are some downsides to this approach because it means that users would need to have `cmdstan` installed before being able to install `epinowcast` whereas at the moment it is optional.

We could also consider keeping the current system but targeting the saved compiled model at the `epinowcast` package as `instantiate` does. I am not sure this would be CRAN compliant though and also not sure it would be that great a user experience.

**Additional context**
Talked a bit about this with @medewitt and anything we decide here is likely of interest to @sbfnk and @jamesmbaazam for `EpiNow2`.

medewitt:
## Description

This PR closes #407 . 

This PR allows the users to set an `enw_cache_location` environment variable. After the initial compilation of the Stan code, as long as the Stan files are the same, calls to `enw_model` will not be recompiled and prior compilations will be used. This PR also adds the ability for the user to set and unset the cache.  To maintain CRAN compliance, the default will always be to use the user's temporary directory.

A new dependency **was not** added for `rappdirs`, but in subsequent vignettes we can recommend this approach as a safe place to cache these models. 

The key functionality is described below:
- set the cache using the `enw_set_cache` function. If nothing is supplied it will default to the `tempdir`. 
- unset the cache using `enw_unset_cache`. This removes the location from the environment, but not from disk, so if they reset the cache to a location that once was a cache, it will still exist.
- a generally internal, but user exposed, `enw_get_cache` allowing to user to see where the cache is (which is written to the .Renviron file). The user could also use the `Sys.setenviron` function if desired (i.e., they don't want to add anything to their dotfiles).

- [ ] I have tested the compilation through different R sessions. I didn't add any tests to ensure this piece as they could greatly extend the CI timelines, but I am happy to do so.

## Checklist


R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#' # Use  the package cache in R >= 4.0
+#' if(R.version.string >= 4.0) {
+#'   enw_set_cache(tools::R_user_dir(package = "epinowcast", "cache"))
+#' }
+#' @importFrom cli cli_abort cli_alert
+#' @export
+
+enw_set_cache <- function(path = NULL) {
+
+  if (is.null(path)) {
```
seabbs, line 350: NULL or missing default path?

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+  }
+
+  prior_cache <- Sys.getenv("enw_cache_location", unset = "", names = NA)
+
+  if (!check_environment_setting(prior_cache)) {
+    cli::cli_alert("{prior_cache} exists and will be overwritten")
+  }
+  env_contents_active <- enw_get_environment_contents()
+
+  #!TODO should we normalise the path?
```
seabbs, line 361: I don't actually have a clue what this means. Explain to me like I'm 5 please!

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#' if(R.version.string >= 4.0) {
+#'   enw_set_cache(tools::R_user_dir(package = "epinowcast", "cache"))
+#' }
+#' @importFrom cli cli_abort cli_alert
+#' @export
+
+enw_set_cache <- function(path = NULL) {
+
+  if (is.null(path)) {
+    cli::cli_abort("`path` must be a valid file path.")
```
seabbs, line 351: this check kind of implies we have checked if it is a valid file path which we haven't. Do you think we should or do anything else here?

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+  )
+
+  writeLines(
+    new_env_contents,
+    con = env_contents_active[["env_path"]], sep = "\n"
+  )
+
+  readRenviron(env_contents_active[["env_path"]])
+
+  invisible(candidate_path)
```
seabbs, line 378: explicit return and no space below to match package style

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#' environment.
+#'
+#' @return the prior cache location, if it existed
+#'
+#' @examples
+#' enw_unset_cache()
+#'
+#' @family utils
+#' @export
+
```
seabbs, line 395: space

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+enw_unset_cache <- function() {
+  prior_location <- Sys.getenv("enw_cache_location")
+
+  Sys.unsetenv("enw_cache_location")
+
+  invisible(enw_get_environment_contents(remove_enw_cache_location = TRUE))
+
+  return(invisible(prior_location))
+
```
seabbs, line 404: space

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+}
+
+
```
seabbs, line 407: space

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#'
+#' @return The string of the filepath
+#'
+#' @return a string representing the file path for the cache location
+#'
+#' @importFrom cli cli_inform
+#'
+#' @family utils
+#' @export
+
```
seabbs, line 423: space

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+enw_get_cache <- function() {
+  cache_location <- Sys.getenv("enw_cache_location")
+
+  if (is.null(cache_location) || cache_location == "") {
+    cache_location <- tempdir()
```
seabbs, line 428: Do you think we should have a cli message here to flag using the temporary directory?

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+  } else {
+    cli::cli_inform("Using `{cache_location}` for the cache location.")
+  }
+
+  return(cache_location)
+
```
seabbs, line 434: space

R/utils.R
```diff
@@ -324,6 +324,145 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#' @family utils
+#' @export
+
+enw_get_cache <- function() {
+  cache_location <- Sys.getenv("enw_cache_location")
+
+  if (is.null(cache_location) || cache_location == "") {
+    cache_location <- tempdir()
+  } else {
+    cli::cli_inform("Using `{cache_location}` for the cache location.")
```
seabbs, line 430: this isa good idea I think but does it mean that the location will be printed with a message here and in enw_model when both are called together (which is fairly common)

tests/testthat/_snaps/enw_set_cache.md
```diff
@@ -0,0 +1,36 @@
...
+      → initial_location exists and will be overwritten
+
+# alert [fancy]
+
+    Code
+      withr::with_envvar(new = c(enw_cache_location = "initial_location"), {
+        enw_set_cache("second_location")
+      })
+    Message
+      → initial_location exists and will be overwritten
```
seabbs, line 35: do we want to explicitly refer to it being the cache location in this message?

tests/testthat/test-enw_get_cache.R
```diff
@@ -0,0 +1,32 @@
+test_that("enw_get_cache can retrieve the enw_cache_location as expected", {
+    withr::with_envvar(new = c(enw_cache_location = "models"),
```
seabbs, line 2: love to see it!

tests/testthat/test-enw_get_cache.R
```diff
@@ -0,0 +1,32 @@
+test_that("enw_get_cache can retrieve the enw_cache_location as expected", {
+    withr::with_envvar(new = c(enw_cache_location = "models"), 
+    expect_equal(
+        basename(enw_get_cache()), 
+        "models"
+    )
+    )
+
+    withr::with_envvar( new = c(enw_cache_location = ""),
+    expect_equal(
```
seabbs, line 10: is being inline with withr:: common practise? It looks a little weird to me?

seabbs (changes requested):
This is really great - colour me excited. 

Left a few/many fussy linting comments. 

Main point is should we tell users about the suggested cache location on load (if its not already set I guess).

R/utils.R
medewitt: Normalisation will expand the path to the fully specified version and will input the platform specific escape patterns (double back slashes on windows) e.g., `normalizePath("..")` -> `C:\\Users\\medewitt\\project\\epinowcast`. I think normalising will result in a safer experience given that Windows can do some wonky things without it. The downside is might be for multi-user systems that run on a generic Renviron file so the cache location is actually different for different users (maybe) or if a user always wants to have a relative path for the cache (e.g., a model directory within their project where the stan code that was used is local to a project). In the latter case they would need to set the cache for each new project or use a local Renviron at the level of the project. Might be overthinking it.

R/utils.R
medewitt: Agree.

R/utils.R
medewitt: Will test this. As you say, maybe it is worth just printing the message `.onLoad` where the cache location is defined (like with cmdstan)

R/utils.R
medewitt: the `cmdstanr::write_stan_file` function is pretty aggressive in creating a directory if it does exist. I could add another message that triggers on a `dir.exists` which could do a "directory does not exist, it will be created" note?

tests/testthat/test-enw_get_cache.R
medewitt: I added at the top of the library calls for `testthat`. To your point, it looks much more clean...

R/utils.R
pearsonca: i think we can use the base R default error for missing arguments work here, rather than setting this to null and checking for null.

then a basic is-it-a-string check before we toss to cmdstanr seems sufficient.

we *could* get into writing a sophisticated directory structure checking / sanitization bit here, but that seems like some other package's job.

R/utils.R
seabbs: @pearsonca comment on this here: https://github.com/epinowcast/epinowcast/pull/416#discussion_r1436600164

medewitt:
Got bogged down on administrative things this week. Outside of some linting and spelling issues very close...

tests/testthat.R
```diff
@@ -1,5 +1,6 @@
 library(testthat)
 library(vdiffr)
+library(widthr)
```
medewitt, line 3: spell this correctly

R/utils.R
```diff
@@ -325,6 +325,135 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#' Stan models rather than a temp directory. This can reduce the
+#' need for model compilation on every new model run across sessions.
+#'
+#' @param path A valid filepath representing the desired cache location
+#'
+#' @return The string of the filepath set
+#' 
+#' @examples
+#' # Set to local directory
+#' dir.create("test")
```
seabbs, line 340: is this a CRAN-compliant example?

R/utils.R
```diff
@@ -325,6 +325,135 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#' enw_set_cache("test")
+#' enw_get_cache()
+#' # Use  the package cache in R >= 4.0
```
seabbs, line 343: double space

R/zzz.R
```diff
@@ -0,0 +1,20 @@
...
+## nocov start
+
+.onLoad <- function(libname, pkgname) {
+    packageStartupMessage(cli_inform(enw_startup_message()))
+}
+
+enw_startup_message <- function(){
+    cache_location <- Sys.getenv("enw_cache_location")
+    if (check_environment_setting(cache_location)) {
+        msg <- "`enw_cache_location` is not set it. Set it using `enw_set_cache` to reduce future Stan compilation times"
```
seabbs, line 12: should we suggest using the R cache here as I think most users won't know about it?

seabbs (commented):
Apart from the linting issues and a question about if we want to tell people when they attached about a suggest cache location then this is looking good to go I think!

R/utils.R
medewitt: Nope :(

medewitt:
@seabbs I think the lintr flag is a false alarm. I was using the conventions in `cli` for the information. How do i get it to ignore the function?

R/zzz.R
```diff
@@ -0,0 +1,30 @@
...
+}
+# nolint end
+
+enw_startup_message <- function() {
+    cache_location <- Sys.getenv("enw_cache_location")
+    if (check_environment_setting(cache_location)) {
+        msg <- c(
+            "!" = "`enw_cache_location` is not set. Set it using `enw_set_cache`
+            to reduce future Stan compilation times.",
+            "i" = "For example: `enw_set_cache(tools::R_user_dir(package =
```
seabbs, line 16: @medewitt agree this looks spurious. We can use # nolint (probably line wrapping here) as discussed here: https://lintr.r-lib.org/reference/exclude.html

seabbs: (also I haven't run this yet but I am excited by how cool this info message looks like it is going to be!)

R/zzz.R
```diff
@@ -0,0 +1,30 @@
...
+
+enw_startup_message <- function() {
+    cache_location <- Sys.getenv("enw_cache_location")
+    if (check_environment_setting(cache_location)) {
+        msg <- c(
+            "!" = "`enw_cache_location` is not set. Set it using `enw_set_cache`
+            to reduce future Stan compilation times.", #nolint
+            "i" = "For example: `enw_set_cache(tools::R_user_dir(package = 
+            \"epinowcast\", \"cache\"))`.", #nolint
+            "i" = "See `?enw_set_cache` for details." #nolint
```
seabbs, line 18:
```suggestion
            "i" = "See `?enw_set_cache` for details." # nolint
```

R/zzz.R
```diff
@@ -0,0 +1,30 @@
...
+# nolint end
+
+enw_startup_message <- function() {
+    cache_location <- Sys.getenv("enw_cache_location")
+    if (check_environment_setting(cache_location)) {
+        msg <- c(
+            "!" = "`enw_cache_location` is not set. Set it using `enw_set_cache`
+            to reduce future Stan compilation times.", #nolint
+            "i" = "For example: `enw_set_cache(tools::R_user_dir(package = 
+            \"epinowcast\", \"cache\"))`.", #nolint
```
seabbs, line 17:
```suggestion
            \"epinowcast\", \"cache\"))`.", # nolint
```

R/zzz.R
```diff
@@ -0,0 +1,30 @@
...
+    packageStartupMessage(cli_inform(enw_startup_message()))
+}
+# nolint end
+
+enw_startup_message <- function() {
+    cache_location <- Sys.getenv("enw_cache_location")
+    if (check_environment_setting(cache_location)) {
+        msg <- c(
+            "!" = "`enw_cache_location` is not set. Set it using `enw_set_cache`
+            to reduce future Stan compilation times.", #nolint
```
seabbs, line 15:
```suggestion
            to reduce future Stan compilation times.", # nolint
```

seabbs:
@medewitt I got a bit excited when reviewing this and added some tests + made the messaging a bit fancier. Let me know what you think and I will also review again before doing anything (in a bit so the fog of excitement has cleared!).

medewitt:
@seabbs looks amazing! Definitely prettier and more informative messaging (and a learning opp for me regarding what you can do with `cli`!).  I'll clean up some of the CI fails in the morning if you don't beat me to them

seabbs:
Time difference helping me win here! I think I got all of the issues + notes + linting problems but we shall see! If you have time to look at this today that would be amazing so we can get it merged in.

seabbs:
I see the following issue when using `devtools::load_all()` in a clean session after having previously used `enw_unset_cache()` in another session. The only place that this could be coming from is the tests so I am very confused? 

```r
devtools::load_all()
ℹ Loading epinowcast
ℹ Using `second_location` for the epinowcast model cache location.
This message is displayed once per session.
```

seabbs:
Hmm I get the following in a clean session with `epinowcast` not even loaded...

```r
r$> Sys.getenv("enw_cache_location")
[1] "second_location"
```

sbfnk:
> Hmm I get the following in a clean session with epinowcast not even loaded...

Presumably because it gets written to your `~/.Renviron` in https://github.com/epinowcast/epinowcast/blob/8b0e4aab3e367246b2e85a5f7266c80cf7d2dceb/R/utils.R#L381

medewitt:
> > Hmm I get the following in a clean session with epinowcast not even loaded...
> 
> Presumably because it gets written to your `~/.Renviron` in
> 
> https://github.com/epinowcast/epinowcast/blob/8b0e4aab3e367246b2e85a5f7266c80cf7d2dceb/R/utils.R#L381

yeah...this is the root cause. We would need to remove `read.environ` and tell the user that they need to restart the session or manually `read.environ`

seabbs:
Ah I see this is because `enw_set_cache` writes to the `.Renviron` but `enw_unset_cache` doesn't

https://github.com/epinowcast/epinowcast/blob/8b0e4aab3e367246b2e85a5f7266c80cf7d2dceb/R/utils.R#L417

Arguably that could be a separate PR but I see we are making progress here!

R/utils.R
```diff
@@ -328,6 +328,192 @@ date_to_numeric_modulus <- function(dt, date_column, timestep) {
...
+#' It can optionally remove the entry for `enw_cache_location`.
+#'
+#' @param remove_enw_cache_location Logical indicating whether to remove the
+#' `enw_cache_location` entry from the `.Renviron` file. Defaults to `TRUE`.
+#'
+#' @return A list containing the contents of the `.Renviron` file and its path.
+#' @keywords internal
+enw_get_environment_contents <- function(remove_enw_cache_location = TRUE) {
+  env_location <- Sys.getenv("HOME")
+  env_path <- file.path(env_location, ".Renviron")
```
seabbs, line 496: What happens if people are using a local/project-specific .Renviron?

R/utils.R
```diff
@@ -420,6 +420,12 @@ enw_unset_cache <- function() {
     ))
     Sys.unsetenv("enw_cache_location")
 
+    clean_environ <- enw_get_environment_contents(
+      remove_enw_cache_location = TRUE
```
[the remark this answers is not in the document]
seabbs: I did this in [`4bb8269` (#416)](https://github.com/epinowcast/epinowcast/pull/416/commits/4bb8269a7476a44873bf27c0937e6cce76174fd3) but the issue is now is that if you unset the non-persistent cache you now can't unset the persistent cache...

R/utils.R
medewitt: Need to fix this. I'll look locally first, then go to `$HOME`

medewitt: https://github.com/epinowcast/epinowcast/pull/416/commits/aa41b0fd120e178012cab9bad5f3e98e0e5f75a2 should fix this piece and testing locally with an empty project with an .Renviron

medewitt:
> Right so I can confirm that current changes resolve the issue and that the `.Renviron` is now updated.
> 
> I am now wondering though if things are getting a little complicated. Do we want to simplify and have a transient and non-transcient cache? Something like this?
> 
> `enw_set_transcient_cache()`, `enw_set_persistent_cache()`, `enw_unset_persistent_cache()`, `enw_unset_transcient_cache()`, `enw_get_cache()`, `enw_get_transcient_cache()`, `ewn_get_persistent_cache()` (where `enw_get_cache()` uses the transcient if present and otherwise the global option?
> 
> Or is that all getting a bit much and isn't simpler so it would be better to stick to arguments?

IMHO we limit the user facing APIs to reduce their cognitive burden and the amount of user facing apis for us to maintain. `enw_get_cache` is simple but will look to the current environment and then go to either the working directory Renviron location (as of a few moments ago) or to the user Renviron. 

We might have to write out the options that a user might want....

```r
enw_unset_cache <- function(pattern = c("all", "session"){
   if ( pattern == "session") { Sys.unsetenv("enw_cache_location")}
  if (pattern == "all") {existing code....}
```

seabbs:
Ah yes I like it and I like the idea of using a character string arg here to allow for more flexibility

seabbs:
@sbfnk would you be up to reviewing this when its ready as I think I might have got a bit stuck in to be objective!

NEWS.md
```diff
@@ -41,6 +41,7 @@ This release is in development and not yet ready for production use.
 - Switched to the `{cli}` package for all package messaging in order to have modern and pretty notifications. See #188 by @nikosbosse and @seabbs reviewed by @pearsonca.
 - Increased the minimum supported R version to >= R 3.6.0 from R 3.5.0 and ensured that existing function code and tests compiled with this dependency. Vignettes will continue to allow use of R >= 4.1.0 syntax (i.e., native pipe and lambda function syntax). See #389 by @medewitt and @seabbs and reviewed by @pearsonca.
 - Add documentation for all custom stan functions. See #422 by @seabbs and reviewed by @sbfnk.
+- Allow for caching Stan models to reduce compilation time through the use of the environment variable, `enw_cache_location`, which can be set using the `set_enw_cache` function. See #407 by @medewitt and reviewed by @seabbs and @pearsonca.
```
seabbs, line 44:
```suggestion
- Allow for caching Stan models to reduce compilation time through the use of the environment variable, `enw_cache_location`, which can be set using the `set_enw_cache` function. See #407 by @medewitt and @seabbs and reviewed by @sbfnk and @pearsonca.
```

sbfnk:
> @sbfnk would you be up to reviewing this when its ready as I think I might have got a bit stuck in to be objective!

Yes.

seabbs:
(I broke it and will fix it - sorry!)

seabbs:
Fixing the tests etc now but I am currently seeing this rather odd behaviour (in both a clean and old cache). 

```r
r$> enw_model(force_recompile = TRUE) -> mod
ℹ Using model /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan/epinowcast.stan.
ℹ Include is /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan.
ℹ Using `/home/seabbs/.cache/R/epinowcast` for the epinowcast model cache location.
Compiling Stan program...
Error in file(con, "r") : cannot open the connection

r$> mod
Error: object 'mod' not found

r$> mod <- enw_model()
ℹ Using model /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan/epinowcast.stan.
ℹ Include is /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan.
ℹ Using `/home/seabbs/.cache/R/epinowcast` for the epinowcast model cache location.
Model executable is up to date!
```

i.e. it errors on compilation but then it has actually compiled when you access the cache.. is this related to this PR? It seems pretty weird to me.

medewitt:
> Fixing the tests etc now but I am currently seeing this rather odd behaviour (in both a clean and old cache).
> 
> ```r
> r$> enw_model(force_recompile = TRUE) -> mod
> ℹ Using model /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan/epinowcast.stan.
> ℹ Include is /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan.
> ℹ Using `/home/seabbs/.cache/R/epinowcast` for the epinowcast model cache location.
> Compiling Stan program...
> Error in file(con, "r") : cannot open the connection
> 
> r$> mod
> Error: object 'mod' not found
> 
> r$> mod <- enw_model()
> ℹ Using model /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan/epinowcast.stan.
> ℹ Include is /home/seabbs/Dropbox/academic/projects/epinowcast/inst/stan.
> ℹ Using `/home/seabbs/.cache/R/epinowcast` for the epinowcast model cache location.
> Model executable is up to date!
> ```
> 
> i.e. it errors on compilation but then it has actually compiled when you access the cache.. is this related to this PR? It seems pretty weird to me. I just updated to the new `cmdstan` version so perhaps that is it?

Seems to work with CmdStan version: 2.32.2

```r
> library(epinowcast)
i Using `C:\Users\medewitt\project\testing-epinowcast\local` for the epinowcast model cache location.
This message is displayed once per session.
> mod <- enw_model()
i Using model C:/Users/medewitt/R/library/epinowcast/stan/epinowcast.stan.
i Include is C:/Users/medewitt/R/library/epinowcast/stan.
i Using `C:\Users\medewitt\project\testing-epinowcast\local` for the epinowcast model cache location.
Compiling Stan program...
> 
```

seabbs:
@medewitt I implemented the character pattern you suggested and updated the testing (and then got very lost in mocking those tests for quite a while!). What do you think? If on board then I think maybe this is ready for review? 

I also moved the exported cache tools into the model tools file as it seemed like a good fit?

Really nice work on this - even from just playing with it it is such a quality of life improvement.

TBH this is such a nice feature it feels like it should be a helper package (along with maybe the new `enw_stan_to_r` functionality) - could spin it out in the epinowcast org. Might be worth an issue I guess.

seabbs (approved):

seabbs:
Hmm failing on windows. Looks like its something to do with how windows is handling paths and that impacts messaging. Also realised there are a bunch of warnings on CI some spurious and some note. For those unrelated to this PR it seems like its its own issue to resolve and maybe for those that are because !!

seabbs (approved):
LGTM (self-review). 

As above I think this looks good from my perspective. The only real gap I can see is that only functionality and not messaging is being tested. We might want to refine this further but perhaps this is good enough for the first version? 

I also think that this should probably be in a helper package at some point (maybe with the stan exposing helper function) but I don't have the bandwidth to set this up at the moment. 

@medewitt great work on this! Do you agree this is in a place it can be merged? @sbfnk a review would be great now I think.

medewitt:
>I also moved the exported cache tools into the model tools file as it seemed like a good fit?

Yes, I agree.

>Really nice work on this - even from just playing with it it is such a quality of life improvement.

Yay! Do you think we should write a vignette on this feature specifically, or adapt some of the other Getting Started/ README docs to show this functionality? 

> As above I think this looks good from my perspective. The only real gap I can see is that only functionality and not messaging is being tested. We might want to refine this further but perhaps this is good enough for the first version?

Agreed. We can tweak these as needed. I'm a bit swamped this week, but we could do some skip on cran testing to hit the coverage points we're missing on the global Renviron.

> I also think that this should probably be in a helper package at some point (maybe with the stan exposing helper function) but I don't have the bandwidth to set this up at the moment.

Agreed. Maybe add to an issue. I could see how we could have a`modelcache` package and could deploy it within `epidist` etc or anything using `cmdstanr` as a backend.

> @medewitt great work on this! Do you agree this is in a place it can be merged? @sbfnk a review would be great now I think.

LGTM. Thanks for making it better!

seabbs:
> Do you think we should write a vignette on this feature specifically, or adapt some of the other Getting Started/ README docs to show this functionality?

I was thinking we would see how it was settling in to see if that was needed. The docs in my view already give quite a nice walkthrough (from the startup message onwards). I was planning on asking someone else (newer to the package) to walk through and see if its all clear and then go based on their opinion?

>  I'm a bit swamped this week, but we could do some skip on cran testing to hit the coverage points we're missing on the global Renviron.

Yeah that is a good idea. Perhaps we can make a new issue and deal with it as and when?

> I could see how we could have amodelcache package and could deploy it within epidist etc or anything using cmdstanr as a backend.

Yes exactly!  Now to find time to write an issue 😆 

Before looking at this I hadn't really grasped how you can use `{cli}` to make the interface output nicer. Looking forward to doing similar stuff across the package now.

sbfnk (approved):
This great! Very much agree that this should be its own package.

Only (minor) comment beyond the individual ones I've left is that the indents are inconsistent in the tests (2/4 spaces) but perhaps that's something we can live with.

R/model-tools.R
```diff
@@ -460,3 +464,178 @@ enw_stan_to_r <- function(
...
+#' if (R.version.string >= "4.0.0") {
+#'  enw_set_cache(
+#'    tools::R_user_dir(package = "epinowcast", "cache"), type = "all"
+#'  )
+#'}
+#'
+#'}
+enw_set_cache <- function(path, type = c("session", "persistent", "all")) {
+
+  type <- rlang::arg_match(type)
```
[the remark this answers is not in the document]
seabbs: its package policy for now to do this with uncommon functions and to just use importFrom with common ones. Is that good practice more generally? I think that is debatable. This certainly follows current style though so I think we should discussion what precisely we want to do in a issue vs here.

R/model-tools.R
```diff
@@ -460,3 +464,178 @@ enw_stan_to_r <- function(
...
+  }
+
+  candidate_path <- normalizePath(path, winslash = "\\", mustWork = FALSE)
+
+  if (type == "persistent" || type == "all") {
+    env_contents_active <- enw_get_environment_contents()
+
+    enw_environment <- paste0("enw_cache_location=\"", candidate_path, "\"\n")
+
+    new_env_contents <- append(
```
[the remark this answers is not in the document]
seabbs: So we could have enw_set_cache(type = "persistent") call enw_unset_cache(type = "persistent") internally?

R/model-tools.R
```diff
@@ -185,7 +185,8 @@ remove_profiling <- function(s) {
 #'
 #' @family modeltools
 write_stan_files_no_profile <- function(stan_file, include_paths = NULL,
-                                        target_dir = tempdir()) {
+                                        target_dir = epinowcast::enw_get_cache()
+                                        ) {
   # remove profiling from main .stan file
```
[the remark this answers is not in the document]
seabbs: yeah that is a good idea. I think perhaps in enw_get_cache or at least a check that it exists.

R/model-tools.R
```diff
@@ -460,3 +464,178 @@ enw_stan_to_r <- function(
...
+      )
+      Sys.unsetenv("enw_cache_location")
+        if (type == "session") {
+          cli::cli_alert_info(
+            "To revert to the persistent cache (if set), run `readRenviron('~/.Renviron')`" # nolint line_length
+          )
+        }
+    } else {
+      cli::cli_alert_danger(
+        "`enw_cache_location` not set in the local environment. Nothing to remove." # nolint line_length
```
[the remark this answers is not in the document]
seabbs: Done in 48157c5f

R/model-tools.R
```diff
@@ -460,3 +464,178 @@ enw_stan_to_r <- function(
...
+  type <- rlang::arg_match(type)
+
+  if (!is.character(path)) {
+    cli::cli_abort("`path` must be a valid file path.")
+  }
+
+  prior_cache <- Sys.getenv("enw_cache_location", unset = "", names = NA)
+
+  if (!check_environment_setting(prior_cache)) {
+    cli::cli_alert_warning("{prior_cache} exists and will be overwritten")
```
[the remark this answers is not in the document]
seabbs: Done in acfaf2d9. Path creation and checking in  7f712a3e

R/model-tools.R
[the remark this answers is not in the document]
sbfnk: Personally I quite like qualifying for clarity and find `@importFrom`s somewhat hard to maintain as they require keeping track of all the uses of an imported function to know when an entry can be removed. But there are a few edge cases where qualifying is really clunky, .e.g. with ``data.table::`[` `` or `data.table::.SD`. Because of this I have ended up sticking with `@importFrom` over qualifying in most instances.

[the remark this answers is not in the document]
seabbs: >  of an imported function to know when an entry can be removed.

Hmm yeah fair. I think making sure they are duplicated so that you have them in every function you use something is pretty key. I have had the reverse with not importing things due to losing track of what the actual imports of the package are (because they aren't in the namespace). 

Agree that `data.table` is a big offender here (along with some others that use less) that is one of the main reasons its entirely imported (and that it is regularly used)

R/model-tools.R
```diff
@@ -460,3 +464,178 @@ enw_stan_to_r <- function(
...
+#' compilation on every new model run across sessions or within a session.
+#' For R version 4.0.0 and above, it's recommended to use the persistent cache
+#' as shown in the example.
+#'
+#' @param path A valid filepath representing the desired cache location.
+#'
+#' @param type A character string specifying the cache type. It can be one of
+#' "session", "persistent", or "all". Default is "session".
+#' "session" sets the cache for the current session, "persistent" writes the
+#' cache location to the user’s `.Renviron` file,  and "all" does both.
```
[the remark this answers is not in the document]
seabbs: I changed some of the wording around and made the recommendation be to use `c("session", "persistent")` rather than `"all"` as this felt more explicit.  As the user is going to copy and paste this felt okay.

seabbs:
Wow this turned into a lot! @medewitt still pretty close to your implementation but with some more safety catches (I hope). What do you think? 

@sbfnk another round of review would be lovely I think I have addressed all of your points aside from the space regex which I found caused problems and I am not sure was needed?

medewitt (commented):
These updates look great~

seabbs:
Thanks for doing the actual content of this! I've just made it complicated I think!

seabbs (approved):

sbfnk (approved):
LGTM

PR #410, opened 2023-12-04
Re #371: moving generic items to organizational level

pearsonca:
## Description

This PR closes #371, by deleting items that have been relocated and updating references to them.

## Checklist

- [X] My PR is based on a package issue and I have explicitly linked it.
- [X] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [X] I have read the [contribution guidelines](https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md).
- [ ] I have tested my changes locally.
- [X] I have added or updated unit tests where necessary.
- [X] I have updated the documentation if required.
- [X] My code follows the established coding standards.
- [ ] I have added a news item linked to this PR.
- [ ] I have reviewed CI checks for this PR and addressed them as far as I am able.

PR #419, opened 2023-12-19
Use enwtheme

Bisaloo:
## Description

This PR solves https://github.com/epinowcast/epinowcast/pull/410#issuecomment-1853958524.

Preview:

Screen Shot 2023-12-19 at 15 49 59-fullpage

pkgdown theme repo: https://github.com/epinowcast/enwtheme

Next step is probably copying the other theming elements there. I'm just waiting for the confirmation we want these elements to be applied across the org first.

https://github.com/epinowcast/epinowcast/blob/cb8434bf90e42c4739345c778bd53c45ef2ce217/_pkgdown.yml#L3-L5

## Checklist

- [ ] I have read the [contribution guidelines](https://github.com/epinowcast/epinowcast/blob/main/CONTRIBUTING.md).


pearsonca:
The upside of this being that as we release other packages, they should also just use this theme and pretty much problem solved? Nice.

So this goes in before #410, then the changes there have no impact on the pkgdown presentation?

Bisaloo:
> The upside of this being that as we release other packages, they should also just use this theme and pretty much problem solved? Nice.

Yes, exactly.

> So this goes in before https://github.com/epinowcast/epinowcast/pull/410, then the changes there have no impact on the pkgdown presentation?

A minor change: I have added a link to the style guide there, which is not present in the standard pkgdown theme. But it seemed to be a good fit so I put it there. Let me know if I should go back to mimicking the standard pkgdown theme as closely as possible.

> aside: seems weird that this would trigger coverage check + benchmarking, right?

It depends what you mean. This matches with what we put in our GHA workflows so everything is happening as planned. It might not be desirable though and it's possible to update our workflow triggers to only run on relevant changes. 
In my experience however, this can cause more frustration because it means that workflows marked as `required` will not always run, and the PR can only be merged by bypassing the branch protection rules.

pearsonca:
> A minor change: I have added a link to the style guide there, which is not present in the standard pkgdown theme. But it seemed to be a good fit so I put it there. Let me know if I should go back to mimicking the standard pkgdown theme as closely as possible.

I think we should have it - @seabbs?
 
> It depends what you mean. This matches with what we put in our GHA workflows so everything is happening as planned. It might not be desirable though and it's possible to update our workflow triggers to only run on relevant changes. In my experience however, this can cause more frustration because it means that workflows marked as `required` will not always run, and the PR can only be merged by bypassing the branch protection rules.

Hmm. To me, the right solution is a bit more like "workflows trigger -> recognize they have nothing to do -> report as such" - keeps the logic of is-change-relevant-to-workflow closer to the actual knowledge. Is that plausible?

seabbs:
This is great - thanks @Bisaloo. 

> A minor change: I have added a link to the style guide there, which is not present in the standard pkgdown theme. But it seemed to be a good fit so I put it there. Let me know if I should go back to mimicking the standard pkgdown theme as closely as possible.

Yes, I agree this is a good thing to have in. 

>Next step is probably copying the other theming elements there. I'm just waiting for the confirmation we want these elements to be applied across the org first.

Yeah, I think we want all of these org wide?

seabbs (dismissed):
Looks great. Nice one @Bisaloo and @pearsonca!

PR #422, opened 2024-01-03
Issue 314: Document stan code

### LINKED ISSUE #314: Document all stan code
At the moment a large number of our stan functions are undocumented. To make the code easier to explore we should change this. 

Unfortunately as far as I am aware there is no documentation system that is supported by stan so we need to adopt our own style here. Sticking to roxygen like documentation probably makes sense given the rest of our code uses this. Alternatively, if there is a supported approach to documentation we should make use of that.

seabbs:
## Description

This PR closes #314.

It adds doxygen style (as suggested by the stan docs) documentation to all custom functions. This aims to be a first pass at what will likely take several iterations so if reviewers find pain points that aren't objectively unhelpful/wrong then I think these should be pushed to new issues. 

Note I have tried to clearly document all dependencies and that there is no way (that I know of) to handle argument inheritance and so there is a lot of duplication.

@adrian-lison pinged you for review as you are probably most across the stan code base and @sbfnk as hopefully having this in place would have improved your contributor journey (which I know was an issue).

See #423 for some thoughts about how to make this documentation easier to parse.

## Checklist


sbfnk (commented):
Great to have these documented!

Upon initial cursory review (happy to do a more detailed one later) there is a lot of repetition here - does doxygen not have any mechanism for inheriting parameters (or could you use `@commonparams` more extensively)?

As an example there are three different definitions of `sl` - the discrepancies are confusing, are they intended?

```
 * @param sl Array of start index for each observation period.
 * @param sl Array of start indexes for each observation period.
 * @param sl Array of start indexes for each snapshot.
```

seabbs:
> does doxygen not have any mechanism for inheriting parameters (or could you use @commonparams more extensively)?

Not that I know of/understand. The stan doc recommendation is really more in the format of doxygen vs actually being doxygen. I think making it work with doxygen probably needs to be part of #423. 

> As an example there are three different definitions of sl - the discrepancies are confusing, are they intended?

No not intended. The perils of duplication. I'll resolve this repetition but I'd really like to address these issues more broadly in follow up issue/PRs.

sbfnk:
> I'll resolve this repetition but I'd really like to address these issues more broadly in follow up issue/PRs.

Would you prefer a more extensive review with suggestions for changes in a separate PR to here?

seabbs:
> Would you prefer a more extensive review with suggestions for changes in a separate PR to here?

Yeah I think a high-level review of is this not misleading/wrong and then a detail review/feedback to go in a follow up issue would be great.

adrian-lison:
Woops, looks like I came just a bit too late :D Does it make sense if I review this now, or should we already make a new PR for detailed improvements?

Really awesome to have this doc in place.

seabbs:
Oops sorry I had it on auto-merge. The perils of asking for a review but only if no one gives any extra work. 

I think now these are in it would be great to go through and flag things that could be improved in some new (ideally modular) issues that are maybe linked by a meta-issue. That way it might open things up a bit to newer contributors etc which might be nice. 

I think we also need to be clear what issues are related to correctness/clarity and what we need to change to get working with Doxygen (or some other generator) (#423)

Issue #307, opened 2023-08-10
rstan exposing functions is failing in CI


seabbs:
As above this is a new issue and is causing all checks to fail. Everything works locally so the precise issue is unclear. I have hotfixed by removing skipping this test on all CI but obviously this isn't a sensible long-term solution. 

See https://github.com/epinowcast/epinowcast/actions/runs/5823160125/job/15789401806

medewitt:
Just sticking this here...CmdStanR now has this functionality (https://mc-stan.org/cmdstanr/reference/model-method-expose_functions.html). Syntax  for handling outputs is slightly different with list environments. I hate rstan deps...

seabbs:
yup we are using that now (sorry should have been clearer when closing)

medewitt:
> yup we are using that now (sorry should have been clearer when closing)

Doh, sorry-- That's what I get when I read things on my phone...

seabbs:
haha no worries!

PR #431, opened 2024-01-10
Issues 268 and 307: Switch out rstan for cmdstanr when exposing stan functions and check in CI

### LINKED ISSUE #268: Switch to using cmdstanr to expose Stan functions to R
Currently we use a conversion to a version of Stan supported by r Stan and then rstan to expose our Stan functions to R for prototyping and user use. 

The dev version of cmdstanr now supports this for the latest version of Stan. 

https://discourse.mc-stan.org/t/exposing-stan-user-defined-functions-using-cmdstanr-and-rcpp/27104/5

We should update to use this functionality and then move to drop out dependence on rstan.

seabbs:
## Description

This PR closes #268 and #307. It should open up much easier stan code testing and potentially more rapid prototyping an simulation code development.

It also:

- Drops our `rstan` dependency
- Renames `extract_stan_functions()` to `enw_stan_to_r()`. As this was not previously exported I have not used any kind of managed depreciation.
- Refactors `enw_stan_to_r()` to use `enw_model()` and to have more automated and user friendly defaults (i.e compiles all package functions it can automatically.
- Refactors `extract_sparse_matrix()` to not use `rstan::extract_spare_parts()`.
- Splits stan functions into additional stan files to make exposing them more modular and to increase the proportion that can be exposed (due to overloading not being supported in R).
- Depreciates converting cmdstan compatible stan code to rstan compatible stan code as this functionality is no longer needed.
- Adds checking exposing stan functions to our cmdstan CI checks.
- `{RcppEigen}` has been added as a suggests as this is required for `{cmdstanr}` to expose stan functions.

Note that exposing stan functions is still a bit unstable and for example fails for me locally when I run `devtools::test()` but not otherwise (with a permissions based segfault).

*Apologies this is a bit of a messy PR - hard to isolate these changes cleanly.*

@sbfnk tagged you for review as you've been quite involved in this part of the code base. If not time etc. flag that and will ping someone else. @medewitt for awareness as this has some impact (in that it uses `enw_model()`) on #416 - feel to review if time!

_Note: This PR is a good example of where committing manual files in PRs gets confusing._

## Checklist


seabbs:
There is an issue with `{touchstone}` benchmarks that looks like an interaction between `{pak}` and `{touchstone}`(trying to install packages not with `{pak}`. This seems unrelated to this PR but could be related to the newly pushed version of `{glue}` on CRAN (this is the package causing the error). I will spin up a test PR to explore this.

R/model-tools.R
```diff
@@ -371,3 +371,92 @@ enw_model <- function(model = system.file(
...
+#' @description This function facilitates the exposure of Stan functions from
+#' the [epinowcast]() package in R. It utilizes the [expose_functions()] method
+#' of [cmdstanr::CmdStanModel] or this purpose. This function is useful for
+#' developers and contributors to the [epinowcast] package, as well as for
+#' users interested in exploring and prototyping with model functionalities.
+#'
+#' @param files A character vector specifying the names of Stan files to be
+#' exposed. These must be in the `include` directory. Defaults to all Stan
+#' files in the `include` directory. Note that the following files are
+#' overloaded and cannot be exposed: "delay_lpmf.stan",
```
sbfnk, line 386:
```suggestion
#' files in the `include` directory. Note that the following files contain
#' overloaded functions and cannot be exposed: "delay_lpmf.stan",
```

sbfnk: Also, is "delay_lpmf.stan" really a problem?

R/model-tools.R
```diff
@@ -371,3 +371,92 @@ enw_model <- function(model = system.file(
...
+#' # or exposed globally and used directly
+#' prob_to_hazard(c(0.5, 0.1, 0.1))
+enw_stan_to_r <- function(
+  files = list.files(include),
+  include = system.file("stan", "functions", package = "epinowcast"),
+  global = TRUE,
+  verbose = TRUE,
+  ...
+) {
+  overloaded_fns <- c(
```
sbfnk, line 421: It's a shame to lose these. You *could* identify them with a tag (in comments) or split by new lines and then identify function names. But perhaps not work the effort / overhead.

sbfnk (commented):
All looks good to me.

sbfnk (dismissed):

R/model-tools.R
seabbs: So identifying them in the comments won't work I think but I think having them have their own names (i.e. blah_blah_vect) would work and may be the way to go but I think perhaps its own issue to resolve?

R/model-tools.R
seabbs: I think so as it depends on functions that can't be exposed.

seabbs:
I won't ping for another review as changes are so minor but feel free to give on if you would like! I will also open another issue for trying to extract the remaining functions.

R/model-tools.R
sbfnk: Sorry I wasn't very clear. You could identify overloaded functions to discard in the comments (or even have an `// *** R ignore from here` or the like which would be easy to filter out when reading in the functions). As for R e.g. you don't need `obs_lpmf(int ...)` as a length 1 vector can be passed to the array version. With this you would then still get a version of all the functions.

PR #432, opened 2024-01-11
Test (DO NOT MERGE): Is touchstone broken

seabbs:
This PR tests if `{touchstone}` breaking is unrelated to #431.

### DIFF
--- a/inst/stan/epinowcast.stan
+++ b/inst/stan/epinowcast.stan
@@ -19,6 +19,7 @@ functions {
 
 data {
   // Indexes and lookups
+  
   int n; // total number of observations (obs)
   int t; // total number of time points 
   int s; // total number of snapshots


sbfnk:
We conclude it's broken? Noting it doesn't seem to be working any more on EpiNow2 either.

seabbs:
> We conclude it's broken? Noting it doesn't seem to be working any more on EpiNow2 either.

Yes and living in hope that it will become unbroken if its just a transient issue. If it is an issue with `{pak}` then potentially it maybe fairly simple to offer a fix to `{touchstone}` but I guess the first thing to do is wait a bit and check if it clears up.

seabbs:
@sbfnk it looks like this was a transient issue. I think ideally (I'm guessing due to the release of glue still) but for now just going to close my eyes and hope it stays away.

PR #446, opened 2024-03-06
444: Add contributors workflow

### LINKED ISSUE #444: Add better visibility of contributors
I think we currently do an okay job surfacing package contributions (in the news and in the DESCRIPTION) but not a great job. Given the nature of this project and the fact that getting credit for this kind of work is so hard already I think its really key we do a great job of this.

I would like an improved solution to this (vague I know) very likely using either the generic `all-contributors` bot or the `Ropensci` version (https://docs.ropensci.org/allcontributors/). I imagine this being either in the README (as standard) or on a custom page (that would then need to be added to the pkgdown site).

I really don't have capacity to set this up and also just don't know what a good solution is so any ideas would be great!

sbfnk:
## Description

This PR closes #444.

It uses the `allcontributors` R package to maintain a section in the README - I chose the "text" format to minimise clutter but other options are available - see [examples](https://docs.ropensci.org/allcontributors/articles/allcontributors.html).

## Checklist

- [X] My PR is based on a package issue and I have explicitly linked it.
- [X] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [X] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [X] I have tested my changes locally.

### DIFF
--- a/.github/workflows/contributors.yaml
+++ b/.github/workflows/contributors.yaml
@@ -5,7 +5,7 @@ on:
 name: Update contributors
 
 jobs:
-  render:
+  update-contributors:
     runs-on: ubuntu-latest
     if: "!contains(github.event.head_commit.message, 'cm-skip')"
     env:
--- a/.github/workflows/contributors.yaml
+++ b/.github/workflows/contributors.yaml
@@ -35,6 +35,7 @@ jobs:
         run: |
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
-          git push https://${{github.actor}}:${{secrets.GITHUB_TOKEN}}@github.com/${{github.repository}}.git HEAD:${{ github.ref }} || echo "No changes to commit"
           git add README.Rmd
           git diff-index --quiet HEAD || git commit -m "Automatic update of contributors"
+          git pull --rebase origin ${{ github.ref.name }}
+          git push origin || echo "No changes to push"
--- a/.github/workflows/contributors.yaml
+++ b/.github/workflows/contributors.yaml
@@ -31,7 +31,7 @@ jobs:
     
       - name: Commit results
         run: |
-          git commit README.* -m 'Update contributors [ci-skip]' || echo "No changes to commit"
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
+          git commit README.* -m 'Update contributors [ci-skip]' || echo "No changes to commit"
           git push https://${{github.actor}}:${{secrets.GITHUB_TOKEN}}@github.com/${{github.repository}}.git HEAD:${{ github.ref }} || echo "No changes to commit"
--- a/.github/workflows/contributors.yaml
+++ b/.github/workflows/contributors.yaml
@@ -29,7 +29,7 @@ jobs:
       - name: Update contributors
         run: allcontributors::add_contributors(format = "text")
         shell: Rscript {0}
-    
+
       - name: Commit results
         run: |
           git config --local user.email "action@github.com"
--- a/.github/workflows/contributors.yaml
+++ b/.github/workflows/contributors.yaml
@@ -35,5 +35,6 @@ jobs:
         run: |
           git config --local user.email "action@github.com"
           git config --local user.name "GitHub Action"
-          git commit README.* -m 'Update contributors [ci-skip]' || echo "No changes to commit"
           git push https://${{github.actor}}:${{secrets.GITHUB_TOKEN}}@github.com/${{github.repository}}.git HEAD:${{ github.ref }} || echo "No changes to commit"
+          git add README.Rmd
+          git diff-index --quiet HEAD || git commit -m "Automatic update of contributors"
--- a/.github/workflows/contributors.yaml
+++ b/.github/workflows/contributors.yaml
@@ -1,0 +1,32 @@
+on:
+  push:
+  workflow_dispatch:
+
+name: Update contributors
+
+jobs:
+  render:
+    runs-on: ubuntu-latest
+    if: "!contains(github.event.head_commit.message, 'cm-skip')"
+    env:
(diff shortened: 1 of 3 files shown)


.github/workflows/contributors.yaml
```diff
@@ -0,0 +1,32 @@
...
+      - uses: actions/checkout@v4
+
+      - uses: r-lib/actions/setup-r@v2
+        with:
+          use-public-rspm: true
+
+      - uses: r-lib/actions/setup-r-dependencies@v2
+        dependencies: NA
+        with:
+          extra-packages: any::allcontributors
```
Bisaloo, line 23:
```suggestion
        with:
          packages: any::allcontributors
```

sbfnk:
Result at https://github.com/epinowcast/epinowcast/blob/contributors/README.md

.github/workflows/contributors.yaml
```diff
@@ -0,0 +1,37 @@
+name: Update contributors
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref }}
+  cancel-in-progress: true
+
+on:
+  push:
```
seabbs, line 8: Do we want this to just trigger on `main`?

seabbs (dismissed):
Love it. This is great - thanks @sbfnk! Only one question.

.github/workflows/contributors.yaml
sbfnk: My thinking was that it increases visibility if it shows up e.g. in pull requests (and might be nice e.g. for someone who opens a new PR to see themselves added).

seabbs: Yeah good point. The downside is that there will be increased bot activity in Prs (readmen and this). Not sure how to weight that up

seabbs:
> Perhaps this could be solved with concurrency groups?
 
 Yes good idea.

sbfnk:
> > Perhaps this could be solved with concurrency groups?
> 
> Yes good idea.

Todo: read up on concurrency groups (unless @Bisaloo has a solution ready).

Bisaloo:
What about making the allcontributors workflow update only `README.Rmd` and then trigger the `render-readme` workflow?

sbfnk:
Yes I think that'll work unless a new contributor updates `README.Rmd`, in which case we'll still get a race / potential push fail in the contributors workflow (though can probably avoid with a condition as above).

Bisaloo:
> Yes I think that'll work unless a new contributor updates `README.Rmd`, in which case we'll still get a race / potential push fail in the contributors workflow (though can probably avoid with a condition as above).

I think it'll still be fine.

1. contributor edits `README.Rmd`
    - This triggers `render-readme`
    - This triggers `allcontributors`
2. `render-readme` pushes new `README.md`
3. `allcontributors` pushes new `README.Rmd` and re-trigger `render-readme`
4. `render-readme` second run pushes new `README.md`

If `allcontributors` finishes first:

1. contributor edits `README.Rmd`
    - This triggers `render-readme`
    - This triggers `allcontributors`
3. `allcontributors` pushes new `README.Rmd` and re-trigger `render-readme`
2. ongoing `render-readme` run is cancelled via the concurrency groups
4. `render-readme` second run pushes new `README.md`

sbfnk:
> 1. contributor edits `README.Rmd`
>    
>    * This triggers `render-readme`
>    * This triggers `allcontributors`
> 2. `render-readme` pushes new `README.md`
> 3. `allcontributors` pushes new `README.Rmd` and re-trigger `render-readme`
> 4. `render-readme` second run pushes new `README.md`

Wouldn't step (3) fail here because local/remote repos differ (assuming they both checked out at roughly the same time)? We could perhaps mitigate by doing a `git fetch` first which would then only fail if the push from (2) happens between initiating fetch and push.

Bisaloo:
> We could perhaps mitigate by doing a git fetch first

Yes, we use `git pull --rebase` before attempting to push in other projects.

sbfnk:
OK, so it seems we have two options:
1. add a dependency to the render-readme action (`needs: contributors`). Downside: will need to set up and maintain forever that any event that triggers `render-readme` also triggers `contributors`. Not clear to me what happens with manual triggering but assuming that this ignores dependencies (could test it in this PR).
2. add a `git pull` before `git push`. Downside: can still break if other action finishes after `git pull` is called and before `git push` finishes

Have I missed anything? Does anyone have any preferences for one or the other?

seabbs:
I think that is a good summary (and thanks @Bisaloo !)

>  Downside: will need to set up and maintain forever that any event that triggers render-readme also triggers contributors

I think I prefer 2. due to this issue and because it makes more sense to me that only one thing ever update the README.md (i.e the render-readme action).

Note that if we have the contributor workflow triggered on PRs as currently we will still get the same issue with hanging CI that we were getting with README rendering.

sbfnk:
That's this issue, right?
https://github.com/epinowcast/epinowcast/issues/447
Was a solution identified before?

seabbs:
Yes. I think in other packages people have moved to having the README just render on `main` - I don't think we found a fix here.

.github/workflows/contributors.yaml
sbfnk: More than one reason to only push on `main` as discussed in #447 so going with that. Done in [c6bc8c4](https://github.com/epinowcast/epinowcast/pull/446/commits/c6bc8c4b26e09478ca8da934b14d3bb82972ee51).

seabbs:
Really nice to have this in!

sbfnk:
Will need to keep an eye on workflows running successfully on `main` as they weren't tested in this PR now.

seabbs:
Damn. I remember a wrinkle now. As we have branch protection on actions can't push to `main`. I've looked into if there is a way around that (i.e giving them enhanced permissions) but I am not sure there is?

Issue #447, opened 2024-03-07
automated workflows that commit don't trigger checks


sbfnk:
In https://github.com/epinowcast/epinowcast/pull/446 the final commit is an automated commit with `[ci skip]` - but if I understand correctly this means merge conditions (5 required checks pass) are never met. Should we avoid using `[ci skip]` in commits from workflows?

Bisaloo:
As discussed in https://github.com/epiforecasts/scoringutils/issues/601, `[ci skip]` is a red herring here and has no functional impact. Commits made by `secrets.GITHUB_TOKEN` will not trigger workflows.

sbfnk:
Thanks, now I remember! So just triggering these on `main` seems indeed the only viable solution.

sbfnk:
Given that this is continuing to be a pain (and the `main` triggering didn't solve things as pushing directly there violates branch protection rules) why don't we use an (organisational-level) PAT instead of the github token?

seabbs:
To get overcome branch protection rules?

sbfnk:
What I meant is: if the fundamental issue is that

> Commits made by `secrets.GITHUB_TOKEN` will not trigger workflows.

why don't we make these commits using a PAT instead, such that workflows to get triggered? Then we could go back to the original plan of updating the README inside PRs (within branch protection rules).

seabbs:
> why don't we make these commits using a PAT instead, such that workflows to get triggered? Then we could go back to the original plan of updating the README inside PRs (within branch protection rules).

Isn't the issue that they don't trigger workflows but the PR expects the workflows to be triggered and so you get stuck unable to merge due to waiting on your required workflows?

sbfnk:
Yes but (if I understand the whole discussion correctly) if we used our own PAT (which would add risk and would have to be appropriately restricted) then actions would be triggered.
That said it failed when trying in
https://github.com/epiforecasts/EpiNow2/pull/609
though it does seem to work in
https://github.com/european-modelling-hubs/covid19-forecast-hub-europe-submissions/blob/32b583511181630c7bd53d8304a613b1220a70d0/.github/workflows/automate_baseline.yml#L44

seabbs:
ah right got you.

sbfnk:
I think this has been closed by #450?

PR #450, opened 2024-04-10
Make Render Readme open a PR vs committing.

seabbs:
## Description

This PR is an attempt to partially address #447 as this is causing some development friction. Rather than continuing to try and commit README etc changes it instead makes the actions open a PR if the action has been triggered on main. 

This comes at a slight cost in that until the README action is merged the md README will be out of date with the Rmd README.

As part of this I have also removed the contributors workflow and put it within the render readme workflow.


seabbs:
Tagged a few people who might have thoughts on if this actually helps or is also annoying!

adrian-lison (approved):
Thanks Sam!

I only looked at the changes but did not test it (how do you test this? :D)

But I like the approach. I would be up for giving this a try.

.github/workflows/render-readme.yaml
```diff
@@ -28,17 +28,34 @@ jobs:
...
-          git add README.md man/figures/
-          git diff-index --quiet HEAD || git commit -m "Automatic readme update [ci skip]"
-          git push origin || echo "No changes to push"
+      - name: Check for README.Rmd changes
+        id: readme_changes
+        uses: dorny/paths-filter@v3
+        with:
+          filters: |
+            readme:
+              - 'README.Rmd'
```
Bisaloo, line 49: I don't believe this step is necessary as this is already handled in `peter-evans/create-pull-request` from what I can see.

Bisaloo (approved):
I agree this change leads to be better workflow.

.github/workflows/render-readme.yaml
seabbs: Ah good point I've removed.

sbfnk (approved):
Looks great!

Issue #400, opened 2023-11-24
Exposing Stan dependency versions to package users


zsusswein:
**Is your feature request related to a problem? Please describe.**

I use `{epinowcast}` in a containerized pipeline where we try to maximize reproducibility. When building new versions of the image, we build `{epinowcast}` from source from this repository pinned to a specific commit. But it can get a bit messy with layer caching and moving versions to get the latest features.

We try to manually keep track of the associated versions of the Stan dependencies and install the correct versions into the container, but it would be a helpful feature to expose those versions programatically. 

**Describe the solution you'd like**
I'm not sure what the most idiomatic solution is, but ideally we'd be able to call a simple function or variable to get the needed information. Something like `epinowcast::get_cmdstan_version()`.

**Describe alternatives you've considered**

Even a simple message to the user that printed on package load would be a nice quality-of-life improvement.

seabbs:
This is a good idea and would also help us in development I think. The only thing we want to be careful of is to not make this overly restrictive as stan is pretty stable so its likely our model won't actually depend on a specific version and we don't want people to install new versions for no reason.

I'm not sure we want to supply a function to do this. We could encode the stan dependency into the DESCRIPTION but that might be overly proscriptive. We could also use the new #386 to add the version of cmdstan to `main` we are use the package work with and then use that in a function to serve the latest recent version?

pearsonca:
> **Is your feature request related to a problem? Please describe.**
> 
> I use `{epinowcast}` in a containerized pipeline where we try to maximize reproducibility. When building new versions of the image, we build `{epinowcast}` from source from this repository pinned to a specific commit. But it can get a bit messy with layer caching and moving versions to get the latest features.

Prying a bit: how are you defining "reproducibility"? In the scientific sense (i.e. prioritizing ability to specify these precise conditions => get these precise results) or in the engineering sense (i.e. on a budget, reliably churning out parts within quality tolerance standards + minimal disruptions to production line with maintenance/upgrades)?

In my opinion, these are antagonistic, so the answer can't be "both". So: what we might actually do to solve the problem depends on which perspective you're arguing to prioritize.

pearsonca:
@zsusswein calling your attention to this discussion as well: https://community.epinowcast.org/t/epinowcast-command-line-interface/207/3, as that seems pertinent to what you're doing. If you could provide example use cases / user stories / etc, the potential outcome of that work would likely be more useful to you et al.

seabbs:
[...]

PR #464, opened 2024-05-16
Add initial support for pathfinder

seabbs:
## Description

This PR adds initial support for Pathfinder. It appears to work fairly well for some models and less well for others (generally based on complexity). Definitely more to do here but I think having this functionality in now is useful.

Some likely contributors who may be interested to review @adrian-lison, @medewitt, @sbfnk.

Something this does not add is pathfinder initialisation for NUTs which I am waiting on support in `cmdstanr` for.

Try it out with: 

```r
library(epinowcast)

pobs <- enw_example("preprocessed")
latest_obs <- enw_example("observations")

path_nowcast <- epinowcast(pobs,
  expectation = enw_expectation(~1, data = pobs),
  fit = enw_fit_opts(enw_pathfinder, pp = TRUE, num_threads = 4, num_paths = 16),
  obs = enw_obs(family = "poisson", data = pobs)
)

plot(path_nowcast, latest_obs)

nuts_nowcast <- epinowcast(pobs,
  expectation = enw_expectation(~1, data = pobs),
  fit = enw_fit_opts(save_warmup = FALSE, pp = TRUE,
	chains = 2, iter_warmup = 1000, iter_sampling = 1000,
  ),
  obs = enw_obs(family = "poisson", data = pobs)
)

plot(nuts_nowcast, latest_obs)
```

@athowes we should check `brms` to see if `Pathfinder` is supported as for the same reasons as its useful here it might be handy to surface in `epidist`. 

## Checklist


R/model-tools.R
```diff
@@ -240,7 +240,7 @@ write_stan_files_no_profile <- function(stan_file, include_paths = NULL,
 #' `model` being used.
 #'
 #' @param model A `cmdstanr` model object as loaded by [enw_model()] or as
-#' supplied by the user.
+#' supplied by th
```
medewitt, line 243: the user got deleted?

DESCRIPTION
```diff
@@ -139,5 +139,5 @@ Encoding: UTF-8
 Language: en-GB
 LazyData: true
 Roxygen: list(markdown = TRUE)
-RoxygenNote: 7.2.3
+RoxygenNote: 7.3.1
 SystemRequirements: CmdStan (>=2.29)
```
medewitt, line 143: I worry about dependencies on an updated CmdStan/ CmdStanR. If users don't run an updated version they might get an esoteric error messaged.

medewitt (commented):
One typo, one philosophical. How do we detect the version of CmdStanR installed such that we can alert the user that pathfinder (and eventually laplace) isn't available? Or do we let CmdStan do it for us?

DESCRIPTION
seabbs: yeah I think this is a good point. We have an issue #400 to try and track the version of `CmdStan` for a specific version of `epinowcast` which feels like it might help?

medewitt: I thought there was issue out there. I think tracking would help for sure. I wonder if we could highjack what `CmdStanR` does right now to test for version and do a quick compare vs the feature we are using? 

See https://github.com/stan-dev/cmdstanr/blob/1f4efadaa3eaf742ebd7e60aa43806b9ae46d119/R/zzz.R#L37C1-L37C21

Issues
- [ ] currently not exported function in `CmdStanR`
- [ ] Would require some maintenance on the `epinowcast` side to ensure that VERSION >= 2.31 or whatnot

seabbs: Yeah that is a good idea. I am a little scared of trying to track minimum deps though honestly. Do you think it would be a massive headache for users to require a specific version (a soft requirement)?

R/model-tools.R
```diff
@@ -289,6 +299,58 @@ enw_sample <- function(data, model = epinowcast::enw_model(),
...
+#'  obs = enw_obs(family = "poisson", data = pobs),
+#' )
+#'
+#' summary(nowcast)
+enw_pathfinder <- function(data, model = epinowcast::enw_model(),
+                           diagnostics = TRUE, ...) {
+  dot_args <- list(...)
+  dot_args$num_threads <- dot_args$threads_per_chain
+  dot_args$threads_per_chain <- NULL
+  fit <- do.call(model$pathfinder, c(list(data), dot_args))
```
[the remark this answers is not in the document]
seabbs: Good idea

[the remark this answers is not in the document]
seabbs: Added in https://github.com/epinowcast/epinowcast/pull/464/commits/68d0330238bcba73b63a6c3483df7fd9d47dd72d

medewitt (approved):
LGTM--can't wait to play with this!

adrian-lison (commented):
Looking good! @seabbs, how much did you already play around with it? I just tested it on your example but with a daily random walk model and generation time, it wasn't too bad in terms of expectation, only the uncertainty quantification got worse (intervals very narrow).

When I tested pathfinder a while ago for `EpiSewer`, I wasn't impressed. It didn't seem to like the renewal model at all... Trends were often considerably off, and there was basically no uncertainty. Interestingly, this got worse when running more paths, results were best when running just one path with a limited number of iterations...

Would be great to get a better understanding of pathfinder's behaviour on time series models and possible tweaks, as this is obviously very useful.

seabbs:
>  on your example but with a daily random walk model and generation time, it wasn't too bad in terms of expectation, only the uncertainty quantification got worse (intervals very narrow).

Yes, I also only did some limited exploration and found similar. I concluded that its likely useful for prototyping but I wouldn't trust it (on the examples I tested) for anything where I wanted to use the output.

@athowes is interested in potentially doing a more in depth look. I think your point about potential tweaks is a good one as it might be something simple that improves performance.

>  Interestingly, this got worse when running more paths, results were best when running just one path with a limited number of iterations...

I didn't see this and that is interesting. Suggests instability based on initialisation?

PR #506, opened 2024-08-21
Update logo usage

seabbs:
To work with newer versions of `pkgdown` this PR uses `usethis::use_logo()`.


seabbs (commented):
I think we need to do this in `EpiNow2` as well (once that has a logo).

seabbs (commented):
LGTM (self-review)

PR #513, opened 2024-08-23
Issue 350: Optimise combine_effects

seabbs:
## Description

This PR is linked to #350 and aims to speed up combining model components. This first pass tries to reduce allocations and the size of matrix multiplications. #514 expanded this PR to add sparse matrix support everywhere.
## Checklist

- [x] My PR is based on a package issue and I have explicitly linked it.
- [x] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [x] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [x] I have tested my changes locally.
- [x] I have added or updated unit tests where necessary.
- [x] I have updated the documentation if required.
- [x] My code follows the established coding standards.
- [x] I have added a news item linked to this PR.
- [x] I have reviewed CI checks for this PR and addressed them as far as I am able.

PR #514, opened 2024-08-25
Issue 350: Stacked PR 2/2 - Sparse design matrices

seabbs:
## Description

This PR adds support for sparse design matrices which should reduce memory and computation for very sparse design matrices. It builds on #513 by adding sparse design matrices as an option. For testing I have left this on but I am undecided if this should be the default. 

Based on benchmarks there appears to be limited benefits in the current test set. In other testing I have also seen limited evidence that this is beneficial. 

This PR also simplifies how we handles intercepts. It drops them everywhere from the design matrix and this allows simplification of the code in `combine_effects` and reduces duplication in the design matrix multiplication.

## Checklist


seabbs:
*`sparse_design = FALSE` is now the default and there is a helper function that checks design matrix sparsity and tells the user about sparse design matrices if their input is very sparse.*

jamesmbaazam:
I'm hoping @pearsonca will take this one, if not, I can take a look next week.

R/check.R
```diff
@@ -671,3 +671,43 @@ check_observation_indicator <- function(
...
+#'
+#' @param matrix A numeric matrix to be checked for sparsity.
+#' @param sparsity_threshold A numeric value between 0 and 1 indicating the
+#' threshold for considering a matrix sparse. Default is 0.9.
+#' @param min_matrix_size An integer indicating the minimum size of the matrix
+#' for which to perform the sparsity check. Default is 50.
+#' @param name A character string specifying the name of the design
+#' matrix. Default is "checked".
+#'
+#' @return This function is used for its side effect of providing an
```
pearsonca, line 688: seems like this should return TRUE or FALSE? i get the use is apparently for the message, but a bit weird to ask a yes-no question and not return the answer. Is there a way for me to programmatically say "if this check indicates sparse, set sprase_design = TRUE"?

inst/stan/functions/utils.stan
```diff
@@ -26,3 +26,31 @@
...
+ * 
+ * @param x The input matrix to be analyzed.
+ * 
+ * @return An integer representing the count of non-zero elements in the matrix.
+ * 
+ * @note This function may be useful in contexts where the sparsity of a matrix
+ * needs to be quantified, such as in sparse matrix operations or in assessing
+ * the efficiency of certain matrix algorithms.
+ */
+int num_nonzero(matrix x) {
```
pearsonca, line 44: bit odd there isn't a built-in sparsity indicator, but doesn't seem to be.

pearsonca (approved):
Seems fine. I'm not quite clear on why (if? could be misreading) its necessary to always include / use / pass around the sparse elements. In my mind, those should only exist if using sparse mode - is that actually what happens?

seabbs:
> sparse elements. In my mind, those should only exist if using sparse mode - is that actually what happens?

yes but you have to pass them around as empty arrays in stan even if they don't exist

R/check.R
seabbs: I don't disagreee but this does fit with the rest of the check funcitons that return NULL

seabbs (commented):
LGTM (self-review)

PR #482, opened 2024-07-26
Issue #481: Add support for probability aggregation to `expected_obs()`

jessalynnsebastian:
## Description

This PR closes #481. It adds an indicator (`agg_probs`) and a matrix argument (`agg_indicator`) to `expected_obs()` that indicate whether to aggregate and how the aggregation should be performed, respectively. The matrix argument is structured to be directly multiplied to the probability vector: the rows represent which index the aggregated probability should be placed on in the new probability vector that accounts for the reporting cycle, and the columns represent whether a given probability in the probability vector should be aggregated. For example, if we had the aggregation matrix 

```math
\texttt{agg}\_\texttt{indicator} = \begin{bmatrix} 0 & 0 & 0 & 0 & 0 & 0 \\\ 0 & 0 & 0 & 0 & 0 & 0 \\\ 1 & 1 & 1 & 0 & 0 & 0 \\\ 0 & 0 & 0 & 0 & 0 & 0 \\\ 0 & 0 & 0 & 0 & 0 & 0 \\\ 0 & 0 & 0 & 1 & 1 & 1 \end{bmatrix}
```
the first 3 elements of the probability vector would be summed and placed on the third element of the new probability vector, and the last three elements would be summed and placed on the sixth element.

For now, `agg_probs` is set to 0 (for no aggregation) and `agg_indicator` is set to [0] in `expected_obs_from_index()`, since the infrastructure to implement this method doesn't yet exist.

## Checklist

- [x] My PR is based on a package issue and I have explicitly linked it.
- [x] I have included the target issue or issues in the PR title in the for Issue(s) *issue-numbers*: PR title
- [x] I have read the [contribution guidelines](https://github.com/epinowcast/.github/blob/main/CONTRIBUTING.md).
- [x] I have tested my changes locally.
- [x] I have added or updated unit tests where necessary.
- [x] I have updated the documentation if required.
- [x] My code follows the established coding standards.
- [x] I have added a news item linked to this PR.
- [x] I have reviewed CI checks for this PR and addressed them as far as I am able.

Issue #480, opened 2024-07-25
Support for fixed reporting schedule via delay probability aggregation


jessalynnsebastian:
This is a sibling issue to #33, which discusses support for a fixed reporting schedule via `-Inf` fixed effects on the hazard.

The use case that motivates this implementation is the case where daily hospital admission nowcasts are desired, and reference dates are available with daily resolution but reporting occurs once per week on a fixed weekday. Preliminary experiments with the approach described in #33 imply that, at least in some cases, it does not do much to improve the stability or performance of the model when compared to fitting a day-of-week reporting pattern.

Another alternative is to pass the reporting schedule into `epinowcast` such that the discretized delay probabilities for each reference date can be aggregated to only the reporting days just before the likelihood.  Any probability of reporting on non-reporting days would be combined and placed on the next possible reporting day.

Some broad steps to implement this:

- [x] #481
  - [x] #489 
- [x] #484
- [x] #483
- [x] #486
- [x] #485

seabbs:
Agree this is a good idea. This looks like a good plan. It might be a good idea to flesh out the remaining tasks in issues asap so we can discuss what would be needed

PR #528, opened 2024-09-10
Issue #484: Add basic probability aggregation capability

### LINKED ISSUE #484: Design input to the Stan code that would link user-supplied input to the aggregation argument to `expected_obs`
I think this is linked to #483 and we may be able to both at once (especially if we leverage the infrastructure from `enw_formula`.

jessalynnsebastian:
## Description

This PR closes #484. It allows the user to input a list of aggregation matrices (see #482) to explicitly model a fixed reporting cycle (see #480).

## Checklist


R/model-modules.R
```diff
@@ -247,6 +241,19 @@ enw_report <- function(non_parametric = ~0, structural = ~0, data) {
     prefix = "rep", drop_intercept = TRUE
   )
 
+  # Check for structural model and define
+  if (as_string_formula(structural) != "~0") {
+    cli::cli_alert_warning(
+      "The structural reporting model is in experimental development"
+    )
+    data_list$agg_probs <- 1
+    # "structural" should be a 2d (nested) list of matricess
```
seabbs, line 250: I'd suggest even at this stage we do a basic check for this. we know it needs to be group by reference time long and we have that information here. we also know it needs to be a D by D matrix

inst/examples/germany_reporting_cycle_model.R
```diff
@@ -0,0 +1,124 @@
...
+)
+
+# Aggregate data to weekly reporting cycle
+repcycle_germany_hosp <- nat_germany_hosp |>
+  enw_add_incidence() |>
+  dplyr::mutate(day_of_week = lubridate::wday(report_date, label = TRUE)) |>
+  dplyr::mutate(confirm = new_confirm) |>
+  # Aggregate rolling sum sums over "confirm"
+  # but we want a sum over "new_confirm"
+  epinowcast:::aggregate_rolling_sum(
```
seabbs, line 36: this is starting to be used a bit in places do you think we should export it?

inst/examples/germany_reporting_cycle_model.R
```diff
@@ -0,0 +1,124 @@
...
+    by = "reference_date"
+  ) |>
+  dplyr::mutate(confirm = ifelse(day_of_week == "Wed",
+                          confirm,
+                          0)) |>
+  dplyr::mutate(not_report_day = ifelse(day_of_week != "Wed",
+                                 1,
+                                 0)) |>
+  # Get confirm as cumulative again
+  dplyr::mutate(confirm = cumsum(confirm), .by = "reference_date")
```
seabbs, line 47: this is fine but there is also a add_cumulative we could be using

inst/examples/germany_reporting_cycle_model.R
```diff
@@ -0,0 +1,124 @@
...
+
+# Define the structural argument for probability aggregation
+ref_t <- unique(pobs$obs[[1]]$reference_date)
+ref_t <- ref_t[!is.na(ref_t)]
+# We are assigning one 35x35 matrix to each reference day, depending on the DOW
+# Report day is Wednesday, and as coded above everything up to & including Wednesday
+# is reported on Wednesday.
+# Scratch in a function to make the agg_indicator for each wday
+generate_agg_indicator <- function(reference_dow_idx, reporting_dow_idx, max_delay) {
+  # Indices are as from lubridate::wday
```
seabbs, line 86: This looks good to me. If it were me I would do it with dataframes though and then cast it out. Having said that I don't see a way to do that which is immediately obvious.

inst/stan/epinowcast.stan
```diff
@@ -118,6 +118,9 @@ data {
   int rep_rncol; 
   matrix[rep_fncol, rep_rncol + 1] rep_rdesign; 
   array[2, 1] real rep_beta_sd_p;
+  // Probability aggregation indicators
+  int agg_probs;
```
seabbs, line 122: we should probably assign this a prefix like repagg_probs and co to match the others?

inst/stan/epinowcast.stan
```diff
@@ -118,6 +118,9 @@ data {
...
+  array[g, s] matrix agg_indicators;
```
seabbs, line 123: So this may have been my confusion but s is I think the number of unique snapshots regardless of group and t is reference time. 

I would expect this to throw a bunch of out of bounds errors for that reason?

I think we have a few options. 1. Use t instead and I think this will work and 2 store this by s (and not g and t) and then in `expected_obs_from_snap` we can use that we are looping over s to pull out the right matrix. Doing that might be a little tricky as you would need to construct the write list structure in the input (or collapse the user supplied one) so perhaps 1 is the way to go

inst/stan/functions/expected_obs_from.stan
```diff
@@ -143,7 +141,7 @@ vector expected_obs_from_snaps(int start, int end, array[] vector imp_obs,
       esnap += l;
       log_exp_obs[ssnap:esnap] = expected_obs_from_index(
         i, imp_obs, rdlurd, srdlh, refp_lh, dpmfs, ref_p, rep_h, ref_as_p, g, t,
-        l, refnp_lh, ref_np, p
+        l, refnp_lh, ref_np, p, agg_probs, agg_indicators
```
seabbs, line 144: See the index lookuip on 133-136 for a sense of the information we have to play with here.

seabbs (commented):
This is looking really nice. I think I have caused some index confusion which I have tried to clarify but otherwise very close.

inst/stan/epinowcast.stan
jessalynnsebastian: Ah, you found this even though I forgot to tag you! Okay, got it, I misunderstood. I think using t instead makes sense.

R/model-modules.R
```diff
@@ -247,6 +241,28 @@ enw_report <- function(non_parametric = ~0, structural = ~0, data) {
...
   )
 
+  # Check for structural model and define
+  if (as_string_formula(structural) != "~0") {
+    cli::cli_alert_warning(
+      "The structural reporting model is in experimental development"
+    )
+    if (!(length(structural) == data$groups)
+      || !(length(structural[[1]]) == data$time)
+      || !(all(dim(structural[[1]][[1]]) == rep(data$max_delay, 2)))) {
```
jessalynnsebastian, line 251: Implemented a basic check here for the format of `structural` but I would have to assume there's a better way to do it

inst/examples/germany_reporting_cycle_model.R
jessalynnsebastian: No strong opinions on this, I think I've only used it to create reporting cycle style data, which I don't think is that common to need to do

inst/stan/functions/expected_obs_from.stan
```diff
@@ -143,7 +142,7 @@ vector expected_obs_from_snaps(int start, int end, array[] vector imp_obs,
       esnap += l;
       log_exp_obs[ssnap:esnap] = expected_obs_from_index(
         i, imp_obs, rdlurd, srdlh, refp_lh, dpmfs, ref_p, rep_h, ref_as_p, g, t,
-        l, refnp_lh, ref_np, p
+        l, refnp_lh, ref_np, p, rep_agg_probs, rep_agg_indicators[g, t][1:l, 1:l]
```
jessalynnsebastian, line 145: added the indexing to expected_obs_from_snaps() here, assuming I just needed to pull it "up" a level from expected_obs_from_index? difficult for me to wrap my head around what to be pulling where

jessalynnsebastian:
@seabbs Addressed some of your comments above. When I try to compile I now have the message that `All elements in list 'rep_agg_indicators' must be numeric!` and wish I could find more documentation on passing a list of lists

inst/stan/functions/expected_obs_from.stan
seabbs: This looks right to me

seabbs:
>  your comments above. When I try to compile I now have the message

When you compile or try and fit? I wouldn't have thought the input data would have an interaction with compiling but just when you try and fit the data?

seabbs:
I've just sight read this so far so will install and try it out

jessalynnsebastian:
Yep, my bad - not when compiling, when fitting a model.

inst/examples/germany_reporting_cycle_model.R
```diff
@@ -105,13 +105,12 @@ generate_agg_indicator <- function(reference_dow_idx, reporting_dow_idx, max_del
...
 ref_t_wdays <- lubridate::wday(ref_t)
-# List has outer length g (1) and inner length t (74 or something)
-agg_indicators <- vector("list", 1L)
-agg_indicators[[1]] <- lapply(seq_len(length(ref_t)), function(t) {
-  return(ref_day_matrix_list[[ref_t_wdays[t]]])
-})
+agg_indicators <- array(dim = c(1, length(ref_t), 35, 35))
+for (t in seq_along(ref_t)) {
+  agg_indicators[1, t, , ] <- ref_day_matrix_list[[ref_t_wdays[t]]]
+}
```
jessalynnsebastian, line 113: @seabbs Formatting agg_indicators like this dealt with that specific error about numeric lists, but now I can't get Stan to initialize when trying to fit a basic model. I'll keep messing with it.

seabbs:
I'll try and give this a try today and see if I can see anything

seabbs:
So I think the remaining issue here is probably in input preparation as the rest of the changes look good to me. I am just writing another version of that preprocessing to see if I can find an issue but so far I haven't seen one (boy this input format is hard to grasp right now). 

The reason I think its this is the error is it sees a -Inf (i.e a 0 prob) that makes me think there is mismatch between the .observed mask and the accumulation matrix. This also means this really can't ever be used where you tried and observe the input so for anything user facing we would need to think about that (as it is super tricky as above)

seabbs:
Once this is passing checks will merge into the dev version

PR #869, opened 2026-06-16
fix(report): derive report axis length from report metadata

### LINKED ISSUE #868: enw_report() recycles rep_findex when the report axis is longer than time + max_delay - 1
Hello, potential bug here identified by Claude. PR coming.

### Describe the bug

`enw_report()` builds the report-date → (group, time) index by reshaping `rep_findex` into a matrix with `nrow = time + max_delay - 1` (`R/model-modules.R`, ~L366–375). That length only holds for the standard reporting triangle. When the report axis has more dates per group — e.g. after `enw_complete_dates(completion_beyond_max_report = TRUE)`, as used to set up a forecast horizon — `length(rep_findex)` no longer divides evenly, so `matrix()` falls back to R's vector recycling. This emits

```
data length [112] is not a sub-multiple or multiple of the number of rows [93]
```

and produces a silently mis-mapped `rep_findex` (report-date effects assigned to the wrong group/time cells). `rep_t` is set from the same assumption. The warning surfaces from `epinowcast()` even with no explicit report module, since `enw_report()` runs with defaults.

### Minimal reprex

```r
library(epinowcast)

obs <- enw_example("observations")
max_delay <- 20

inc <- enw_complete_dates(
  obs,
  max_delay = max_delay,
  max_date = max(obs$report_date) + 14,   # extend the report axis (forecast setup)
  completion_beyond_max_report = TRUE
)
pobs <- enw_preprocess_data(inc, max_delay = max_delay)

# report dates per group exceed time + max_delay - 1:
nrow(pobs$metareport[[1]]) / pobs$groups[[1]]   # 112
pobs$time[[1]] + max_delay - 1                  # 93

enw_report(~0, data = pobs)   # Warning: "...not a sub-multiple or multiple..."
```

### Expected behaviour

`enw_report()` should use the actual number of report dates per group (from `metareport`) so the index maps correctly regardless of report-axis length — or error clearly if the report metadata is not rectangular across groups.

### Environment

- epinowcast 0.6.0.1000 (`main`, commit `b7a2bfec`)
- R 4.5, macOS (arm64); backend-independent (R-side reshape)

nicholasdavies:
### Summary

Fixes #868. `enw_report()` assumed the per-group report axis was `time + max_delay - 1` and reshaped `rep_findex` against it. When the report axis is longer (e.g. `enw_complete_dates(completion_beyond_max_report = TRUE)`), `matrix()` silently recycled `rep_findex` — emitting a "not a sub-multiple" warning and mis-mapping report-date effects.

### Changes

- `enw_report()` now takes the report dates per group from the report metadata (`nrow(metareport) %/% groups`) for both the `rep_findex` reshape and `rep_t`, and errors if the report metadata is not rectangular across groups.
- Regression test in `tests/testthat/test-enw_report.R` covering an extended report axis: asserts no recycling warning and correct `rep_t`.
- NEWS entry.

In the standard reporting triangle `nrow(metareport)/groups == time + max_delay - 1`, so existing fits are unchanged; only the previously-broken extended-axis case changes behaviour.

### Reprex (before: warns; after: clean)

```r
obs <- enw_example("observations")
inc <- enw_complete_dates(obs, max_delay = 20,
  max_date = max(obs$report_date) + 14, completion_beyond_max_report = TRUE)
pobs <- enw_preprocess_data(inc, max_delay = 20)
enw_report(~0, data = pobs)   # no longer warns; rep_t == 112
```

### Checklist

- [x] Regression test added
- [x] `NEWS.md` updated
- [x] `devtools::test()` passes locally (only failures are pre-existing `test-plot.*` snapshot tests needing the optional `vdiffr` package, unrelated to this change)
- [x] `lintr` clean on the changed code

## Bug Fixes

* Fixed `enw_report()` incorrectly recycling the report-date index when the report axis extends beyond the standard time + max_delay - 1 calculation. Report-date effects are now correctly mapped using actual report metadata, preventing cross-group/time effect mapping errors and eliminating spurious data length warnings.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -51,6 +51,10 @@
   The fixed-effects design matrix was collapsing to a vector after the intercept was dropped, causing Stan to error with `mismatch in number dimensions declared and found in context`.
   See #783 by @seabbs.
 
+- Fixed `enw_report()` recycling the report-date index (`rep_findex`) when the report axis is longer than `time + max_delay - 1`, for example after `enw_complete_dates(completion_beyond_max_report = TRUE)`.
+  Previously this emitted a "data length is not a sub-multiple or multiple" warning and mis-mapped report-date effects across groups and times; the number of report dates per group is now read from the report metadata.
+  See #868.
+
 # epinowcast 0.6.0
 
 This release prepares the package for CRAN submission and introduces new methods for inspecting `epinowcast` and preprocessed data objects, including `print()`, `summary()`, `plot()` and an `enw_get_data()` accessor.
--- a/R/model-modules.R
+++ b/R/model-modules.R
@@ -363,16 +363,14 @@ enw_report <- function(non_parametric = ~0, structural = NULL, data) {
   }
 
   # map report date effects to groups and times
+  rep_t <- nrow(data$metareport[[1]]) %/% data$groups[[1]]
+  if (rep_t * data$groups[[1]] != nrow(data$metareport[[1]])) {
+    cli::cli_abort("Report metadata is not rectangular across groups.")
+  }
   data_list$rep_findex <- t(
-    matrix(
-      data_list$rep_findex,
-      ncol = data$groups[[1]],
-      nrow = data$time[[1]] +
-        data$max_delay - 1
-    )
+    matrix(data_list$rep_findex, ncol = data$groups[[1]], nrow = rep_t)
   )
-  data_list$rep_t <- data$time[[1]] +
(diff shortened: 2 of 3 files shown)


seabbs (approved):
Looks good to me. Thanks for this fix and original report @nicholasdavies!

seabbs:
Failing CI is just a spelling false positive which will get swept up on the next spell check scheduled PR so merging.

